PR #3, opened 2018-02-15
fix #2 empty .launcher.conf

### LINKED ISSUE #2: error for empty .launcher.conf (+ hard to debug)
In Quantum Mobile, launching the mc-home app for the first time writes an empty `.launcher.conf` into the app directory.
This causes the app to crash, when one runs it a **second** time.

Since the `.conf` file is executed via `eval`, the error message is rather cryptic for python standards:
```
File "<string>"
SyntaxError: unexpected EOF while parsing
```

I'll make a pull request for this in a bit.

ltalirz:
Fix #2 

This changes to a json format for the config file,
which eliminates the dirty `eval` statement.

Since json can not represent sets, 2 lines of code are added
for hide/show (code is removed in other places).
I think this is a fair tradeoff.

### DIFF
--- a/start.ipynb
+++ b/start.ipynb
@@ -20,7 +20,8 @@
     "from pprint import pformat\n",
     "from markdown import markdown\n",
     "from importlib import import_module\n",
-    "from IPython.display import clear_output, Markdown"
+    "from IPython.display import clear_output, Markdown\n",
+    "import json"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


ltalirz:
@oschuett By the way, please let me know whether you like to sign off on things like this or whether you are fine with us proceeding in our own.

oschuett:
Unfortunately, I won't have the time to review everything in depth anymore. So, please go ahead and make the apps your own.

I'll continue to maintain appmode and aiida-cp2k for the time being.

PR #5, opened 2018-02-20
A simple update feature for the app manager.

edditler:


ltalirz:
Thanks @edditler .
I've had a brief look now and I think there is still potential for simplification, I'll do a proper review later when I have time.

Just a few suggestions
* I would replace the app-level "check update" button by an "update" button. Clicking this button will bring the app up to date (informing the user whether something changed or not)
* please use python3-compatible print statements as everywhere else

Cheers,
Leo

edditler:
Thanks for your thoughts.

> I would replace the app-level "check update" button by an "update" button. Clicking this button will bring the app up to date (informing the user whether something changed or not)

I am not a pro in things concerning git, but my idea for separating the `git fetch` and `git pull` was to avoid running into conflicts when the local version of an app has changed and a merge might be necessary to complete the `git pull`.

If we disregard this potential caveat, should the "Check all for updates" button also turn into "Update all"?

ltalirz:
>  my idea for separating the git fetch and git pull was to avoid running into conflicts when the local version of an app has changed and a merge might be necessary to complete the git pull.

In my view, if we detect that a user has edited the app (in whatever way),  we should not be updating automatically.

Of course, you can still help them along by displaying the full set of commands they would need to perform on the console like `cd ...; git ...` but I would not to it for them.

oschuett:
For merging notebooks the [nbdime](https://github.com/jupyter/nbdime) tools might be useful.

appmanager.ipynb
```diff
@@ -76,12 +76,17 @@
     "        if app_select.value:\n",
     "            display_app_box(app_select.value)\n",
     "\n",
-    "app_select = ipw.Select()\n",
-    "update_app_list()\n",
+    "app_select = ipw.Select(layout={'height': '{}px'.format(len(registry)*20)})\n",
     "app_select.observe(on_app_change, names='value')\n",
```
ltalirz: Would `overflow_y: 'visible'` work instead of computing the height?
http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Styling.html#Flexbox

appmanager.ipynb
```diff
@@ -76,12 +76,17 @@
...
     "\n",
     "out_app = ipw.Output()\n",
-    "display(app_select, out_app)"
+    "out_update = ipw.Output()\n",
```
ltalirz: Can you reuse the other output?
Currently, your output stays, even if one selects another app afterwards.

appmanager.ipynb
```diff
@@ -103,12 +108,112 @@
...
+    "        ver_sel = mk_version_selector(name) # move this down so the UI is more responsive\n",
+    "        ref_drop.options=ver_sel['options']\n",
+    "        ref_drop.value=ver_sel['head']\n",
+    "        def on_ref_change(c):\n",
+    "            change_ref(name, ref_drop.value)\n",
+    "        ref_drop.observe(on_ref_change, names='value')\n",
+    "        \n",
+    "    try:\n",
+    "        btn_remove.disabled = name == \"home\"\n",
+    "    except UnboundLocalError:\n",
```
ltalirz: When/why is this UnboundLocalError thrown?

appmanager.ipynb
edditler: The options don't overflow the container so setting `overflow_y: visible` has no effect  and the `size` attribute is not exposed by the widget so we can't adjust it.

https://stackoverflow.com/questions/9943886/make-multiple-select-to-adjust-its-height-to-fit-options-without-scroll-bar

appmanager.ipynb
edditler: ✓

appmanager.ipynb
edditler: It should not be thrown at all now.

appmanager.ipynb
```diff
@@ -76,12 +76,17 @@
...
-    "app_select = ipw.Select()\n",
-    "update_app_list()\n",
+    "app_select = ipw.Select(layout={'height': '{}px'.format(len(registry)*20)})\n",
     "app_select.observe(on_app_change, names='value')\n",
     "\n",
     "out_app = ipw.Output()\n",
-    "display(app_select, out_app)"
+    "out_update = ipw.Output()\n",
+    "\n",
+    "btn_check_all = ipw.Button(description='Check all for updates.')\n",
```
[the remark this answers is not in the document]
edditler: ✓

edditler:
I implemented the changes and now check that the commits are ancestors of each other by
```
    git_head_hash = ! git -C $appdir rev-parse HEAD
    git_latest_hash = ! git -C $appdir rev-parse origin/master
    # returns ['0'] if the latest commit is an ancestor of HEAD
    git_is_ancestor = !git -C $appdir merge-base --is-ancestor $git_head_hash $git_latest_hash; echo $?
```

ltalirz:
Hi @edditler, very sorry for letting your PR hang in here for so long. We are now looking into this again.
I just checked it out and:
 * when I make a local commit, it still updates automatically
 * when I have local changes of the files it still updates automatically
It's important to avoid this because either of the two points above make it impossibly hard for us to debug problems of users - they will think they just updated to the latest version of the app and they wonder why it is not working.

Fyi, @yakutovicha will work on this later today and perhaps make a few changes in this respect.

PR #31, opened 2020-02-11
Warn if the AiiDA version is lower than 1.0

yakutovicha:


yakutovicha:
> I was thinking that we would update the docker image and recommend users to restart the server and update the whole setup instead of just the home application.

Not all AiiDA lab users run AiiDA lab under a docker container. Some of them use Quantum Mobile virtual image. That is why I think the home app is the best place to put the update check. All interactions with AiiDA lab start from it.

csadorf:
> > I was thinking that we would update the docker image and recommend users to restart the server and update the whole setup instead of just the home application.
> 
> Not all AiiDA lab users run AiiDA lab under a docker container. Some of them use Quantum Mobile virtual image. That is why I think the home app is the best place to put the update check. All interactions with AiiDA lab start from it.

So this would trigger updating `aiidalab` through the dependencies of the home app? That could work I think.

yakutovicha:
The check I added is to detect that currently installed aiida is lower than 1.0.0 - which means the system is old. So the button changes the home app branch to 'support/aiida-0.12.x'.

Would you like to pass by to see how it works in practice?

csadorf:
After offline discussion, the only change I would suggest is not completely remove the original repository but instead move it to something like `$ORIGINAL_DIR-backup-$DATETIME`.

start.ipynb
```diff
@@ -74,15 +75,17 @@
     "    def update_home(_=None):\n",
     "        home = os.path.expanduser('~')\n",
     "        home_app_path = os.path.join(home, 'apps', 'home')\n",
+    "        home_app_new_path = home_app_path + '-backup-' + str(datetime.date.today())\n",
```
csadorf:
```suggestion
    "        home_app_new_path = home_app_path + '-backup-' + datetime.now().isoformat())\n",
```

yakutovicha: fine with me :)

yakutovicha:
@csadorf one more approval, please 😺

PR #38, opened 2020-08-11
Move all widgets from aiidalab to home

csadorf:


csadorf:
This version was tested with both `aiidalab:develop` and `aiidalab:csadorf:refactor/move-widgets-to-home-app`.

PR #39, opened 2020-08-11
(WIP) Prepare home app for app-environments feature.

csadorf:


csadorf:
Testing this locally, the appstore fails with a `TypeError`.
Screen Shot 2020-08-11 at 7 52 08 PM
I recommend that we implement a test that would catch this issue before moving forward.

PR #44, opened 2020-08-20
Add test for the terminal

dou-du:


csadorf:
I'll review this.

tests/test_start.py
```diff
@@ -15,6 +15,7 @@ def get_new_windows(selenium, timeout=2):
 def test_click_appstore(selenium, url):
     selenium.get(url("apps/apps/home/start.ipynb"))
     with get_new_windows(selenium) as handles:
+        time.sleep(5)
```
csadorf, line 18: The sleep is not at the right spot. It should be after clicking on the puzzle piece to give aiidalab enough time to collect some apps and display them. You can see that none of the apps are shown in the appstore screenshot.

tests/test_terminal.py
```diff
@@ -0,0 +1,18 @@
...
+    selenium.get(url("apps/apps/home/start.ipynb"))
+    selenium.set_window_size(1575, 907)
+    selenium.find_element(By.CSS_SELECTOR, ".fa-terminal").click()
+    page = selenium.window_handles[-1]
+    sleep(3)
+    selenium.switch_to.window(page)
+    selenium.find_element(By.CSS_SELECTOR, ".xterm-cursor-layer").click()
+    selenium.find_element(By.CSS_SELECTOR, ".xterm-helper-textarea").send_keys("jupyter --version") 
+    selenium.find_element(By.CSS_SELECTOR, ".xterm-helper-textarea").send_keys(Keys.ENTER) 
+    sleep(1)
```
csadorf, line 17: Any chance you could verify the appearance of the text after hitting enter?

csadorf:
@dou-du Were you able to figure out what the issue might be?

dou-du:
> @dou-du Were you able to figure out what the issue might be?

So fart, I do not have any idea. One question, is it possible to login the aiidalab/aiidalab-test-app-action after testing ? I try it but I cannot restart it after testing.

csadorf:
> > @dou-du Were you able to figure out what the issue might be?
> 
> So fart, I do not have any idea. One question, is it possible to login the aiidalab/aiidalab-test-app-action after testing ? I try it but I cannot restart it after testing.

If you look at the [`run_tests.sh`](https://github.com/aiidalab/aiidalab-test-app-action/blob/master/run-tests.sh) script of the aiidalab-test-app-action, you see that it automatically shuts down and removes the test environment, however you can either manually edit the file to skip that or we could extend the script to make that optional.

dou-du:
Hi @csadorf , finally I figure out the problem is that we need pause several minus when clicking the "apply filter" button. Now, I obtained the fully loaded screenshot. Please review and merge this pull request. Thanks.

csadorf (approved):
@dou-du Thank you for implementing these tests and resolving this issue. It is somewhat annoying that we need to rely so much on these manual timeouts. I think we should strive to rely more on the "[waits](https://selenium-python.readthedocs.io/waits.html)" in future iterations.

PR #47, opened 2020-10-16
Feature/app compatibility with environment

csadorf:
This is a large update to the home app to support the features introduced in https://github.com/aiidalab/aiidalab/pull/129 and https://github.com/aiidalab/aiidalab/pull/134 and some more cosmetic updates to the user interface.

Additional notes:
- It is easiest to test this update in combination with https://github.com/aiidalab/aiidalab/pull/133 to be able to directly modify the app registry.
- Tests are expected to fail until aforementioned PRs for aiidalab are merged.

### DIFF
--- a/home/widgets.py
+++ b/home/widgets.py
@@ -80,7 +80,7 @@ class AppStatusInfoWidget(ipw.HTML):
             f'{Theme.ICONS.APP_DETACHED} Modified</font></div>'
 
     TOOLTIP_APP_INCOMPATIBLE = \
-            "None of the available app versions support the current AiiDAlab environment. "\
+            "None of the available app versions are compatible with this AiiDAlab environment. "\
             "You can continue using this app, but be advised that you might encounter "\
             "compatibility issues."
 
@@ -89,8 +89,8 @@ class AppStatusInfoWidget(ipw.HTML):
             f'{Theme.ICONS.APP_INCOMPATIBLE} App incompatible</font></div>'
 
     TOOLTIP_APP_VERSION_INCOMPATIBLE = \
-            'The currently installed version of this app is not supported for this '\
-            'AiiDAlab environment. Click on &quot;Manage App&quot; to install a supported version '\
+            'The currently installed version of this app is not compatible with this '\
+            'AiiDAlab environment. Click on &quot;Manage App&quot; to install a compatible version '\
             'and avoid compatibility isssues.'
 
     MESSAGE_APP_VERSION_INCOMPATIBLE = \
--- a/home/widgets.py
+++ b/home/widgets.py
@@ -59,7 +59,7 @@ class StatusHTML(_StatusWidgetMixin, ipw.HTML):
         self.show_temporary_message(change['new'])
 
 
-class UpdateAvailableInfoWidget(ipw.HTML):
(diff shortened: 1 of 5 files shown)


csadorf:
@yakutovicha I'll leave this PR in draft stage until the changes related to aiidalab are merged, but this is ready for testing.

csadorf:
Merge after: https://github.com/aiidalab/aiidalab/pull/134

yakutovicha:
@csadorf something is wrong with the tests. I tried to rerun them, and they failed.

csadorf:
> @csadorf something is wrong with the tests. I tried to rerun them, and they failed. Can you have a look?

I don't think those tests can succeed before the aiidalab version on the docker stack is updated.

yakutovicha:
> > @csadorf something is wrong with the tests. I tried to rerun them, and they failed. Can you have a look?
> 
> I don't think those tests can succeed before the aiidalab version on the docker stack is updated.

ok

home/widgets.py
```diff
@@ -59,28 +61,73 @@ def _observe_message(self, change):
         self.show_temporary_message(change['new'])
 
 
-class UpdateAvailableInfoWidget(ipw.HTML):
+class AppStatusInfoWidget(ipw.HTML):
     """Widget that indicates whether an update is available."""
```
yakutovicha, line 65: Maybe change the docstring to better reflect the changes.

yakutovicha (commented):
I think the code looks good. I also tested the behaviour and didn't notice any issues. I left one minor comment to update the docstring. Otherwise, provided that the tests are fixed - we can merge.

csadorf:
@yakutovicha The tests still fail, because the selenium test for the app management widget fails. That's actually kind of expected, because the behavior of that widget has slightly changed (good sign that the test caught that 👍 ). That said, I've tried to re-implement the same test and had some trouble getting something proper, because I found no straight-forward way to detect whether an app is actually installed or uninstalled via selenium. It's obvious to a human, because the label texts change, but I have not yet found a way to access that information via selenium.

yakutovicha:
> @yakutovicha The tests still fail, because the selenium test for the app management widget fails. That's actually kind of expected, because the behavior of that widget has slightly changed (good sign that the test caught that 👍 ). That said, I've tried to re-implement the same test and had some trouble getting something proper, because I found no straight-forward way to detect whether an app is actually installed or uninstalled via selenium. It's obvious to a human, because the label texts change, but I have not yet found a way to access that information via selenium.

Maybe count how many accordeons are there? Or how many app titles. They belong to a specific CSS class, so shouldn't be difficult to find how many they are:

Screenshot 2020-10-27 at 20 38 32

csadorf:
@yakutovicha I've simplified the unit test and it now passes. It should still give us an indication if there is some issue with the app management widget.

I spent some time on trying to detect whether the button color changed (via css selection) and for some reason that only worked on my desktop. Your proposed solution is a possible option, but would be even more complicated with multiple windows etc. I'm calling it for now, because I've already spent a little too much time on this.

home/widgets.py
csadorf: Done.

yakutovicha (approved):
Thanks, @csadorf, looks good!

PR #55, opened 2021-02-03
Remove unnecessary widget state update after app update.

csadorf:
Fixes aiidalab issue #145.


csadorf:
@yakutovicha I believe this should fix https://github.com/aiidalab/aiidalab/issues/145 .

All tests are failing, but there appears to be an issue with the runner (docker-compose fails to start). I'll investigate, but this appears unrelated to the actual fix here.

yakutovicha (approved):
Looks good, thanks a lot @csadorf !

PR #62, opened 2021-07-14
Package cleanup

csadorf:
Add pre-commit checks, refactor the code accordingly and add a setup.cfg file with updated dependencies.


yakutovicha (approved):
Looks good, thanks @csadorf

PR #69, opened 2021-09-14
Remove the "Update required" state which is now obsolete.

yakutovicha:


csadorf (changes requested):
The `updates_available` trait is also used in the `home/app_manager.py` and `home/start_page.py` modules. Can you make sure to remove the use of this state there as well?

yakutovicha:
> The `updates_available` trait is also used in the `home/app_manager.py` and `home/start_page.py` modules. Can you make sure to remove the use of this state there as well?

I am a bit confused here. I believe we want to keep the "Updates available" state. I only removed the "Updates required".

I think the use of the `updates_available` is now correct.

csadorf (approved):

PR #73, opened 2021-10-05
Import start.py using absolute path.

yakutovicha:
Previously, the apps folder was considered a python module. Since
this is not true anymore [1,2], the loading of the apps' start.py
needs to be adapted.

[1] https://github.com/aiidateam/aiida-prerequisites/pull/37
[2] https://github.com/aiidalab/aiidalab-docker-stack/pull/213

### DIFF
--- a/home/utils.py
+++ b/home/utils.py
@@ -1,7 +1,7 @@
 """Helpful utilities for the AiiDAlab tools."""
 
 import sys
-from importlib import util
+from importlib.machinery import SourceFileLoader
 from os import path
 
 import ipywidgets as ipw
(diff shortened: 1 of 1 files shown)


yakutovicha:
> Looks good. Could you briefly comment about the motivation to use this particular approach as opposed to `importlib.machinery.SourceFileLoader`? I did some cursory search and found the following article: https://www.geeksforgeeks.org/how-to-import-a-python-module-given-the-full-path/

I didn't find the `importlib.machinery.SourceFileLoader` approach, before. But since it is shorter, I will use it.

yakutovicha:
changed, tested - works.

PR #75, opened 2021-10-13
AppStore: fix selecting apps according to their categories.

yakutovicha:
To be tested in combination with https://github.com/aiidalab/aiidalab/pull/240

### DIFF
--- a/home/app_store.py
+++ b/home/app_store.py
@@ -110,7 +110,7 @@ class AiidaLabAppStore(ipw.HBox):
         """This function creates a list of apps to be displayed. Moreover, it creates a parallel list of categories.
         After this the page selector update is called."""
         self.apps_to_display = [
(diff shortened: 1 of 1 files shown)


yakutovicha:
> The `AiidaLabApp` class should not be directly instantiated, but _always_ instantiated via the `from_id()` constructor method. What is the motivation for this patch?

But `from_id()` is gone, no? At least this is the error I got.

csadorf:
> Why then do we do:
> 
> https://github.com/aiidalab/aiidalab-home/blob/84bed162a21b466f7c59a66e95f116b52a9852ea/home/app_store.py#L75-L77
> 
> ?

You're right, calling the constructor like this calls the `from_id()` function internally on the data class: https://github.com/aiidalab/aiidalab/blob/c439d3e901179a7ce3f113c50e28bfb24abc0faf/aiidalab/app.py#L571-L572

PR #96, opened 2022-07-07
Implement open_app tool.

csadorf:
The tool allows to reliably open an app by ensuring that is installed
before opening it.

### DIFF
--- a/open_app.ipynb
+++ b/open_app.ipynb
@@ -1,0 +1,132 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


csadorf:
Video demo of the flow when a particular app is _not_ already installed:

https://user-images.githubusercontent.com/1441208/177804585-30472569-91cf-4928-b79d-f3cc690bc844.mp4

In case that it was already installed, the tool would just redirect immediately.

@CasperWA does that suite your needs?

PR #121, opened 2022-11-24
Fix version mismatch

danielhollas:
In the last release, the version was not actually updated. 

This didn't broke anything, because this package is not published anywhere, and is installed in aiidalab-docker-stack directly from the git repo based on the git version tag, which was correct. But if you try to identify the version in the current image
```
mamba list aiidalab-home
```
it will print the old version.


danielhollas:
@unkcpz would you mind giving me access to this repo as well? :blush:

unkcpz:
> @unkcpz would you mind giving me access to this repo as well? 😊

Sure, done! You gradually have your hands on all repos 😄

unkcpz (approved):
Looks good to me. But why it is not updated? The variable is in the bumpver setup of `setup.cfg`.

danielhollas:
The last release was done by hand, the bumpver was introduced only in the last commit if you look at the git history.

unkcpz:
I see, thanks for check!

danielhollas:
Btw: I caught this by writing a new test as part of https://github.com/aiidalab/aiidalab-docker-stack/pull/327 :-)

PR #128, opened 2023-02-07
Strict dependencies check for aiida-core

unkcpz:
fixes https://github.com/aiidalab/aiidalab/issues/318

The compatibilities only checked for the installed app. If the user going to install the app that would potentially override the aiida-core version the installation will fail in AppManager but the user is not notified before they trigger the installation procedure. 
In this PR, the check for the to-be-installed app is activated when the version is selected and will block the user to install the app that has incompatible "core" dependencies. The "core" dependencies now are only `aiida-core`.
The method `_check_strict_dependency` is used to check that the `aiida-core` version of the container is not overridden by the app. 

If the "core" dependencies are not incompatible, the dependencies to be installed or updated will show in the info notes to remind the user what will be overridden.


unkcpz:
To test with this PR, you can clone the repo and ln the `/home/jovyan/app/home` to the new folder.

home/app_manager.py
```diff
@@ -234,17 +253,64 @@ def _refresh_prereleases(self, change):
...
+
+        if app_version is None:
+            return []
+
+        dependencies = []
+        unmatched_requirements = self.find_unmatched_requirements(
+            app_version, python_bin
+        )
+        packages = find_installed_packages(python_bin)
+        for requriment in unmatched_requirements:
```
danielhollas, line 275: typo, needs to be fixed below as well
```suggestion
        for requirement in unmatched_requirements:
```

home/app_manager.py
```diff
@@ -234,17 +253,64 @@ def _refresh_prereleases(self, change):
...
+
+    def find_to_be_installed_dependencies(self, app_version, python_bin=None):
+        """return a list of Dependency"""
+        if python_bin is None:
+            python_bin = sys.executable
+
+        if app_version is None:
+            return []
+
+        dependencies = []
```
danielhollas, line 270: Clearer naming?
```suggestion
        dependencies_to_install = []
```

home/app_manager.py
```diff
@@ -234,17 +253,64 @@ def _refresh_prereleases(self, change):
...
+        return dependencies
+
+    def _check_strict_dependencies(self, dependencies):
+        """Check dependencies that are not allowed to be override
+        return True if the strict dependencies satisfied
+
+        Note: Only check aiida-core compatibility at the moment
+        """
+        for dep in dependencies:
+            if "aiida-core" in dep.installed:
```
danielhollas, line 293: Please generalize this code for the future, define a list of strict dependencies as a global constant (probably at the top of the file or somwhere).

I also wonder, should this function be here or rather in `aiidalab` repo? I don't have an answer now, but something to think about.

home/app_manager.py
```diff
@@ -234,17 +253,64 @@ def _refresh_prereleases(self, change):
...
+            version_to_install = self.version_selector.version_to_install
+            dependencies = self.find_to_be_installed_dependencies(
+                version_to_install.value
+            )
             compatible = len(self.app.available_versions) > 0
             registered = self.app.remote_update_status is not AppStatus.NOT_REGISTERED
             cannot_reach_registry = (
                 self.app.remote_update_status is AppStatus.CANNOT_REACH_REGISTRY
             )
+            strict_dependency = self._check_strict_dependencies(dependencies)
```
danielhollas, line 313:
```suggestion
            strict_dependency_violation = self._check_strict_dependencies(dependencies)
```
or maybe
```suggestion
            violates_strict_dependency = self._check_strict_dependencies(dependencies)
```

home/app_manager.py
```diff
@@ -234,17 +253,64 @@ def _refresh_prereleases(self, change):
...
+        return True
+
     def _refresh_widget_state(self, _=None):
         """Refresh the widget to reflect the current state of the app."""
         with self.hold_trait_notifications():
             # Collect information about app state.
             installed = self.app.is_installed()
             installed_version = self.app.installed_version
+            version_to_install = self.version_selector.version_to_install
+            dependencies = self.find_to_be_installed_dependencies(
```
danielhollas, line 305:
```suggestion
            dependencies_to_install = self.find_to_be_installed_dependencies(
```

home/app_manager.py
```diff
@@ -396,14 +462,24 @@ def _refresh_widget_state(self, _=None):
...
                     app=self.app
                 )
             else:
                 self.compatibility_info.value = ""
 
+            # Check and show the dependencies install infos
+            if not busy and can_switch:
+                self.dependencies_install_info.value = (
+                    self.DEPENDENCIES_INSTALL_INFO.render(
+                        dependencies=dependencies,
```
danielhollas, line 477:
```suggestion
                        dependencies=dependencies_to_install,
```

home/app_manager.py
```diff
@@ -90,6 +95,18 @@ class AppManagerWidget(ipw.VBox):
         </div>"""
     )
 
+    DEPENDENCIES_INSTALL_INFO = Template(
+        """<div class="alert alert-info alert-dismissible">
+        <a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>
+            Following dependencie(s) will be changed:
+            <ul>
+            {% for p in dependencies %}
```
danielhollas, line 103:
```suggestion
            {% for p in dependencies_to_install %}
```

danielhollas (changes requested):
Hi @unkcpz, this is great, thanks for working on this. I did not test yet, for now just a couple of clarifying questions and some suggestions for variable names.

1. Does this mean that no changes are needed in the `aiidalab` repo? 
2. Will this also allow us to solve the spurious "Update available", if the new version is incompatible? Perhaps that's for a follow up PR?
3. We will need to backport this to the old docker stack image. We need to find out if the current `aiidalab-home` is still compatible or if there were any breaking changes in the meantime.

unkcpz:
I address the variables naming suggested by @danielhollas and close this one with replaced it by #130.

PR #130, opened 2023-02-08
Add strict validation check for core dependencies

unkcpz:
Replacement of #128 

along with https://github.com/aiidalab/aiidalab/pull/347 will fixes https://github.com/aiidalab/aiidalab/issues/318

The compatibilities only checked for the installed app. If the user going to install the app that would potentially override the aiida-core version the installation will fail in AppManager but the user is not notified before they trigger the installation procedure.
In this PR, the check for the to-be-installed app is activated when the version is selected and will block the user to install the app that has incompatible "core" dependencies. The "core" dependencies now are only aiida-core.
The method `validate_strict_dependencies` is used to check that the aiida-core version of the container is not overridden by the app.

If the "core" dependencies are not incompatible, the dependencies to be installed or updated will show in the info notes to remind the user what will be overridden.

- [x] after merging and relase https://github.com/aiidalab/aiidalab/pull/347


home/app_manager.py
```diff
@@ -90,6 +92,24 @@ class AppManagerWidget(ipw.VBox):
...
+    CORE_DEPENDENCIES_COMPATIBILTIY_WARNING = Template(
+        """<div class="alert alert-danger">
+    The version to be install of this app is not compatible with this AiiDAlab environment.
+    </div>"""
+    )
+
+    DEPENDENCIES_INSTALL_INFO = Template(
+        """<div class="alert alert-info alert-dismissible">
+        <a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>
+            Following dependencie(s) will be changed:
```
yakutovicha, line 104:
```suggestion
            The following dependencie(s) will be changed:
```

home/app_manager.py
```diff
@@ -90,6 +92,24 @@ class AppManagerWidget(ipw.VBox):
         </div>"""
     )
 
+    CORE_DEPENDENCIES_COMPATIBILTIY_WARNING = Template(
+        """<div class="alert alert-danger">
+    The version to be install of this app is not compatible with this AiiDAlab environment.
```
yakutovicha, line 97: Maybe like this is a bit better.
```suggestion
    The to-be-installed version of this app is not compatible with this AiiDAlab environment.
```

yakutovicha (changes requested):
Thanks, @unkcpz, I tested the combination of aiidalab-home and aiidalab PR - they seem to work! I didn't have time yet to closely look at the code, but what I wonder is - does it make sense to show the incompatible versions? Maybe we should hide them and have a tickbox to show the incompatible ones only per user request?

Otherwise, I have no quick way to pick up a version that works for my current setup. I need to enumerate all of them to find out which version is compatible. It might be a bit tedious.

Also, I propose to change the messages a bit.

home/app_manager.py
danielhollas: "dependecie" is not a word so putting (s) in parentheses looks super weird to me. I'd just go with `dependencies` even if there might be just one.

```suggestion
            The following dependencies will be changed:
```

home/app_manager.py
```diff
@@ -173,14 +185,17 @@ def __init__(self, app, with_version_selector=False):
             (self.version_selector.installed_version, "value"),
             transform=self._formatted_version,
         )
+        ipw.dlink(
+            (self.version_selector.version_to_install, "value"),
+            (self.app, "version_to_install"),
```
danielhollas, line 190: Why is this needed?

danielhollas (changes requested):
@unkcpz thanks so much. I don't have time to fully test this today, I'll try on Monday.
For now just a couple of minor comments on the code.

home/app_manager.py
unkcpz: Because the dataclass `_AiiDALabApp` is for store the information of the installed app and it is "Static" in terms of App it self. It has no class variable `version_to_install` in this base dataclass. From my understanding, the `version_to_install` is the information given and changed by the widget which wraps the `AiiDALabApp` traitlets. So here I using `dlink` to make `version_to_install` of `self.app` changed passively depend on the `self.version_selector.version_to_install`.

home/app_manager.py
unkcpz: I add a commit https://github.com/aiidalab/aiidalab-home/pull/130/commits/7a87dca3654c236dd926a2171caa561ac49db8b6 to address this comment. But I find it can be simplified by what I mentioned having a more versatible `Dependency` dataclass, then I do not need to do the core package compatible check in the widget here but simply check with the property of the Dependency object.

yakutovicha:
After some thinking, I would propose:

1. Do not list versions that have incompatible core dependencies (a user cannot do much about them, since the installation is forbidden anyways)
2. If no version that is compatible with core dependencies found, then show a banner: "No version compatible with the current environment found". (only display relevant information)
3. Only show the updates list when entering `single_app.ipynb` and not in the app store. (the app store output should be minimalistic as much as possible).

unkcpz:
I fully agree with these, especially the first I proposed once. Please go ahead with it.

danielhollas:
@yakutovicha I agree with what you wrote and what we just discussed. Please go ahead and once a new version is ready I'll test it. This will be a great enhancement.

yakutovicha:
ready for testing @unkcpz and @danielhollas

unkcpz:
thanks a lot for the updates, I did a test and looks much better. Without look at code, I think one thing can improve is when the the http://localhost:8892/apps/apps/home/single_app.ipynb?app=quantum-espresso page loading, there flash out a red alert box and then quickly disappear. I think it comes from when the traitlets not set the box show but then quickly the widget get refreshed and all fine. Do you think it is easy to fix? You can reproduce it by refreshing the singe_app.ipynb of QeApp.

unkcpz (commented):
Looks all good to me, thanks @yakutovicha!
Only the minor issue I mentioned with the animation I record. BTW, I cannot approve this since I open the PR.

unkcpz:
Besides the issue I mentioned above which I'll have a look today to see if it is easy to fix, I think this PR need to be merged after we merge the https://github.com/aiidalab/aiidalab/pulls and make a new release.

danielhollas:
@unkcpz please ping me when / if you manage to solve the issue, and I will test both this and the other PRs together. Then I agree they should be merged, new version of `aiidalab` and `aiidalab-home` are released and we make a new docker stack release.

unkcpz:
@danielhollas I fix the issue the alert popped in first load, which is introduced from `aiidalab` the validate method. Please see the fix at https://github.com/aiidalab/aiidalab/pull/347/commits/ce0c0d239ba9c5679324fa70c48d1c29cdcffd8d. 
It is ready to review I think.

danielhollas (changes requested):
The code itself looks good to me, but I would like to cleanup the UI a bit. 

Specifically, the appstore app list currently contains a lot of unneeded noise.

1. The "include pre-releases" toggle should be hidden. If the user want a pre-release, they should go to the "manage app" page anyway. 
2. On the other hand, I don't like that the version that would be installed by clicking the install button is not shown. I'd like for it to be visible so that I know what I am installing.
Something like "Latest version: XX" Above the install/uninstall buttons.
3. "The manage app" link should be hidden for apps that are not compatible with the environment.

4. The install / uninstall / update controls should be hidden for apps that are not compatible.

5. It is a bit unclear, whether "The app is not compatible with this enviroment" belongs to the app above or below it. If the previous point is done, I'd propose to move it below the title and description instead of the install/uninstall buttons. It seems to me more logical to first read what the app is and then read the warning.

6. To the point above, the warning should say something like No version of the SSSP Toolbox App compatible with the current environment has been found".

image

7. The footer that says "the app is not supported in the current environment" with the "Ignore" ticker should not be displayed.

If some of these require non-trivial work, feel free to push it to another PR. Same comments apply to points that do not strictly pertain to the disabled apps, feel free to create a separate issue and separate PR. I just wanted to get the overall UI feedback here.

unkcpz:
> IMHO the "include pre-releases" toggle should be hidden. If the user want a pre-release, they should probably go to the "manage app" page anyway. (but this is a minor point not related to this PR)

I don't fully agree with this, we have this functionality and it benefits for the user when need to keep it here. In essence, AiiDAlab encourages quick development so the pre-release should be common. I open issue https://github.com/aiidalab/aiidalab-home/issues/138 for further discussion or future implementing.

> On the other hand, I don't like that the version that would be installed by clicking the install button is not shown. I'd like for it to be visible so that I know what I am installing.
Something like "Latest version: XX" Above the install/uninstall buttons.

Addressed by https://github.com/aiidalab/aiidalab-home/pull/130/commits/7576ec500868d2baebd746ae636e11ef7f5c8585, it is the ready to have feature, the version to be installed is in the install button. I think it is a but that only the installed app shows the version to be installed previously.

> The install / uninstall / update controls should be hidden for apps that are not compatible.

I think this is not a good idea, as a user look at the app store, I want to see that the app shown in the list have paralleled layout. If something is not shown in an app, would make me fill uncomfortable. 
If we look at other GUI package manager such as ubuntu software manage and apple store, the install button is always activated to allow users to install but will pop the incompatible information after the button is clicked if the version is not compatible. I think current implementation is quite similar. We cann't pop out the dialog so we deactivate the button for click.

> It is a bit unclear, whether "The app is not compatible with this enviroment" belongs to the app above or below it. If the previous point is done, I'd propose to move it below the title and description instead of the install/uninstall buttons. It seems to me more logical to first read what the app is and then read the warning.

I have the same feeling about this, but instead might be more straightforward to separate by line. I try to add a border to the app manager widget. However, it looks very ugly. If you check carefully, there is already a horizontal line separate widgets. Not sure where it is implement, we can make it more obvious. But I implement the next point, which also clarify the case.

image

> The footer that says "the app is not supported in the current environment" with the "Ignore" ticker should not be displayed.

Addressed by https://github.com/aiidalab/aiidalab-home/pull/130/commits/989262d5c60461069f95c388b8c7d90f0c536b3c by moving out the compatible not satisfied case out and turn off the ignore checkbox.

danielhollas (changes requested):
@unkcpz thanks for the changes! We're getting very close. There's a bunch of general UI improvements that could still be done but we can do later.

One last thing I noticed, when you load the single app page, the Install button flashes several times before all the versions are loaded. I think the logic needs to be improved a bit, the button should only be enabled when all available versions are loaded. (for example, in the very beginning the button gets enabled with version "n/a"). (side note, the "refresh_state" function is crazy complicated now, would be good to add a bunch of tests and split it up in the future).

If this turns out to be too much more work, please just open an issue so we can go ahead with this.

home/app_manager.py
```diff
@@ -191,7 +202,31 @@ def __init__(self, app, with_version_selector=False):
...
+                ]
+            ),
+            ipw.HBox([self.install_info]),
+            ipw.HBox([self.dependencies_log]),
+            ipw.HBox([self.issue_indicator, self.blocked_ignore]),
+        ]
+        if not minimalistic:
+            children.extend(
+                [
+                    ipw.HBox([self.compatibility_info]),
```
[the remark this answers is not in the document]
unkcpz: I don't know. But since it is working at the moment, I don't want to concern about it at the moment.

unkcpz:
@danielhollas thanks!

>  I think the logic needs to be improved a bit, the button should only be enabled when all available versions are loaded. (for example, in the very beginning the button gets enabled with version "n/a"). (side note, the "refresh_state" function is crazy complicated now, would be good to add a bunch of tests and split it up in the future).

During validating the version, the widget should be in the `busy` status. We have a `_show_busy` contextmanager to use.

home/app_manager.py
```diff
@@ -87,6 +103,20 @@ class AppManagerWidget(ipw.VBox):
                 </li>
             {% endfor %}
             </ul>
+
+            The compatibility of dependencies may break by other apps, you can uninstall and install this app again to fix it
```
unkcpz, line 107: I add a note here to show what to do if the incompatibility happened. Usually this caused by installing another package that has different dependencies and break the compatibility of this app.

danielhollas: Really great idea! :clap: I suggest minor rewording. We need to make clear that only uninstalling the app does not help, because the dependencies will not be rolled back to the previous state. The user needs to:
1. Uninstall the app that broke the environment.
2. Reinstall the broken app.
 
```suggestion
            WARNING: Reinstalling previously installed dependencies may break already installed apps. If that happens, uninstall this app and reinstall the app that was broken.
```

(You may want to adda `<br>` linebreak to make this nicer if necessary, I haven't tried)

When an incompatibility happens, pip actually prints a warning at the end. Also `pip check` will detect issues. I will open a separate issue for this, we should run `pip check` at the end of app install and surface any problems to the user.

danielhollas: Oh, I misunderstood where this message is. It's in the template of the app that is incompatible. The sentence above should go to the `DEPENDENCIES_INSTALL_INFO`. Here I would say

```suggestion
            The compatibility issues may be caused by other installed apps. To solve this, you can re-install this app, and uninstall the app that caused the conflict.
```

home/app_manager.py
```diff
@@ -87,6 +103,20 @@ class AppManagerWidget(ipw.VBox):
...
+
+    DEPENDENCIES_INSTALL_INFO = Template(
+        """<div class="alert alert-info alert-dismissible">
+        <a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>
+            The following dependencies will be changed:
+            <ul>
+            {% for p in dependencies_to_install %}
+                <li> {{ p.installed if p.installed is not none else '[Not Installed]' }} --> {{ p.required }} </li>
+            {% endfor %}
+            </ul>
```
danielhollas, line 119: Per comment above, I would add the following
```suggestion
            </ul>
            
            WARNING: Reinstalling previously installed dependencies may break already installed apps. If that happens, uninstall this app and reinstall the app that was broken.
```

danielhollas (changes requested):
> During validating the version, the widget should be in the busy status. We have a _show_busy context manager to use.

Okay, great. Shall we use it then? I don't see any such change in the last two commits.

Otherwise I have only two wording suggestions.

unkcpz:
> Okay, great. Shall we use it then? I don't see any such change in the last two commits.

Ah, the change is in `aiidalab` PR, forget to push there. I'll do it tomorrow, the change is on my office workstation 😭

home/app_manager.py
unkcpz: Yes, there is. It shows when the warning box with details of which dependencies are not matched. This box only shown when the "App incompatible" is shown on the home page.
Changed.

home/app_manager.py
unkcpz: Okay, I like the warning, but the warning in the blue alert-info box may be strange :p I changed it, and let's show  demo next week see others' opinion.

danielhollas (commented):
Thanks @unkcpz. All good. I'll approve this once the `aiidalab` PR is ready and I can do the final testing.

danielhollas:
@unkcpz I noticed an extra "Ignore" checkbox in the appstore for incompatible apps. Also, this checkbox would flash at the beginning when loading the "Manage app" page. I've fixed it in the last commit, please take a look.

obrazek

(I also changed the wording of warnings a little bit)

danielhollas (approved):
Let's go! :tada: 

Great work on this @unkcpz @yakutovicha, this is a great feature.

yakutovicha (approved):
I will also approve to satisfy github 😆

unkcpz:
Cheers! 💪 thanks both!

PR #135, opened 2023-02-20
Remove not depracated git_setup.ipynb

unkcpz:
@yakutovicha This notebook is not used anymore, correct?


danielhollas:
Hmm, this actually looks kinda useful. Perhaps this could be a part of the new "Settings" tab that we've discussed as part of making AWB into a library.

unkcpz:
Why do you think it is useful? I think it basically provides for AiiDAlab user git access but in my opinion, the "production" AiiDAlab user doesn't need this and we (developers) need git setup but since we introduce the aiidalab-launch to start the local development environment this again not been used.

danielhollas:
Yeah, I guess for production this shouldn't be there.

danielhollas (approved):
LGTM but please wait what @yakutovicha has to say before we delete this.

yakutovicha (approved):
oh, yeah. That is something extreeeeemely old and not used anymore.

PR #136, opened 2023-02-21
Home page: Open external URL in a new tab

### LINKED ISSUE #124: URL button should open a new tab
I believe it would be nicer if the following "URL" button opened the App's URL in a new tab.
image

danielhollas:
Because the home page loads so slowly, it's extra annoying if you navigate away by accident. I make ore sense to open Apps' URL in a new tab. Same for "Manage App" button.

https://github.com/aiidalab/aiidalab-home/pull/new/open-url-new-tab

Closes #124

### DIFF
--- a/home/start_page.py
+++ b/home/start_page.py
@@ -159,11 +159,11 @@ class AppWidget(ipw.VBox):
             header_items.append(app_status_info)
 
             footer_items.append(
-                f"<a href=./single_app.ipynb?app={app.name}><button>Manage App</button></a>"
+                f"""<a href=./single_app.ipynb?app={app.name} target="_blank"><button>Manage App</button></a>"""
             )
             if app.metadata.get("external_url"):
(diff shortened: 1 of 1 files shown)


unkcpz (approved):
Looks good to me, and I think open in a new page makes sense for this case.

unkcpz:
I don't know if the initial design is to try to keep all app working in the open tabs as possible to make it more like app? I'll wait @yakutovicha to double-check. But at list for this case, we are move from home to app manager or external url, which should be regarded as the "another" app.

danielhollas:
Yeah. For me the main issue is that the loading is so slow so navigating in between pages in a single tab is very annoying.

(I was thinking about if we could make the home page static so that it loads quickly, but that's a discussion for another time).

yakutovicha (approved):
I think this change is perfectly valid.

PR #140, opened 2023-02-22
Enlarge output from a running installation

### LINKED ISSUE #139: UI: Improve rolling output when installing in AppStore
The output from pip when installing from AppStore only shows 5 lines, which makes it quite hard to see what is happening. Also, the output does not automatically roll down while the installation is ongoing. 

obrazek

The output disappears after the installation is complete, although I would personally prefer if it stayed there even after the successfully install in case one wants to inspect the details (in the case we need to make it very obvious that the installation finished successfully). But here opinions can differ. :-)

danielhollas:
Enlarging the console output from a running installation to make it easier to see what is happening. Closes #139

OLD:

image

NEW:
image

### DIFF
--- a/home/app_manager.py
+++ b/home/app_manager.py
@@ -5,11 +5,11 @@ from subprocess import CalledProcessError
 
 import ipywidgets as ipw
 import traitlets
+from aiidalab.app import AppRemoteUpdateStatus as AppStatus
+from aiidalab.app import AppVersion
 from jinja2 import Template
 from packaging.version import parse
 
-from aiidalab.app import AppRemoteUpdateStatus as AppStatus
-from aiidalab.app import AppVersion
 from home.utils import load_logo
 from home.widgets import LogOutputWidget, Spinner, StatusHTML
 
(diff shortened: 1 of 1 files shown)


home/app_manager.py
```diff
@@ -410,7 +410,9 @@ def _show_msg_success(self, msg):
 
     def _show_msg_failure(self, msg):
         """Show a message indicating failure to execute a requested operation."""
-        self.install_info.show_temporary_message(HTML_MSG_FAILURE.format(msg))
+        # DH: Setting the value directly so that the error message
+        # does not disappear
+        self.install_info.value = HTML_MSG_FAILURE.format(msg)
```
danielhollas, line 415: I hate when error messages just disappear. That's not a good design.

unkcpz: Just a quick question on this, if change the version and retrigger the installation, the output will be wiped and logged from scratch or append to the existing content? 
For the increase in the output panel part, I 100% agree.

Issue #133, opened 2023-02-18
Disable the confirmation dialog for static pages


danielhollas:
Currently, whenever you try to navigate away from an appmode app (e.g. by clicking a URL link, or closing a tab), I almost always get a confirmation dialog, even if I didn't really interact with the page. This gets especially annoying for the home app (e.g. every time I want to close it, I need to confirm).

I would propose to disable this dialog for pages that are static, such as the home page, appstore and perhaps others.

I've tested that the following code in a jupyter code cell achieves the effect.

```js
%%javascript
window.onbeforeunload = function() {return}
```

The original dialog seems to come from jupyter notebook, see file `/opt/conda/lib/python3.9/site-packages/notebook/static/notebook/js/notebook.js`

I've verified that even with the override, the python kernel gets killed when the app window is closed. It seems that this is handled by the appmode via the event listener here:

https://github.com/oschuett/appmode/blob/master/appmode/static/main.js#L26

Of course this would need to be tested more carefully.

@unkcpz @yakutovicha LMK your thoughts on this, happy to take this on if you agree, once we're done with the current changes.

unkcpz:
Hi @danielhollas, thanks for bringing this out. I also once took some time looking at this and I think we should not turn the notification off for all notebooks. 
What I understand is the dialog popped out because the appmode will create a copied notebook ipynb file with the index as a suffix to run the notebook and show the widgets. After the new notebook run, it is changed and so being asked to save the changes. 
If the functionality is turned off globally, for the regular notebook the user won't see the notification if the cell is changed and not saved, this is a useful reminder to ask the user to store the changes in case they forget about it. 

I think maybe in the jupyterkernel can introduce a rule to recognise for some specific names of the "created on the fly" notebook from appmode and only for these notebooks we turn off the popped out dialog.

danielhollas:
Hm, not sure if messing with jupyterkernel is a good idea. I think the logical place to put this is Appmode, if we want to disable it for all apps.

For now as a stop gap solution, I would at least add the JS snippet to the Home app, because that is the most annoying one with which the user always interacts, and the pop-up is sooo annoying every time I want to close the tab.

unkcpz:
> the user always interacts, and the pop-up is sooo annoying every time I want to close the tab.

Couldn't agree more. Please go ahead with the home app. If that works well we can move on and try more static notebooks.

yakutovicha:
[...]

PR #141, opened 2023-02-22
Disable prompt dialog when exiting home page

danielhollas:
This part of a solution for #133 (see more info there). 

In this PR, I am only affecting the home page. If this proves to work, we can apply the same for other pages, such as the appstore or the "Manage app" page.

Essentially, I just added a piece of Javascript that disables the prompt functionality, coming from Jupyter notebook. I've put the cell purposefully at the end of the notebook so that it is executed last. When a developer would be working on this notebook, the alert will not be disabled unless they execute the last cell.

Testing plan:
1. Open home page
2. Open "Tasks". You should see the running home page notebook.
3. Close the home page. There should be no prompt warning from the browser.
4. Verify that the notebook disappears from the task list.

@unkcpz Can you give it a try? It would be great to have this merge for the next release so we can test it out more broadly.


danielhollas:
I just had an idea how to improve this. We should be able to detect inside the javascript, whether we're running in the Appmode or not, which would make this workaround safe. Marking as a draft for now.

danielhollas:
@unkcpz @yakutovicha This should be ready. Please take a look and test (I suggested the testing plan in the OP). Let me know if the suggested commit message makes things sufficiently clear.

unkcpz:
It turns out the solution is quite simple. I am not very familiar with javascript so need a close look after it. I give it a test and it works fine. I can now change the apps order in the home page without seeing the annoying prompt dialog, and the change of app order also gets saved after refresh.

unkcpz (approved):

danielhollas:
Thanks @unkcpz.

Note that I've applied the same fix in my app and have been using this for last two weeks and did not notice any issues.

@yakutovicha If you agree I'll merge this so we can give this a try.

yakutovicha (approved):
I tested old and new versions. I agree with the change 👍

PR #146, opened 2023-03-09
Suppress should scroll for single app manager

unkcpz:


unkcpz:
I don't know why this is only happening now. I think previously it is all fine.

unkcpz:
I cannot reproduce it on my laptop, will try tomorrow on my workstation.

danielhollas:
I just experienced this when I added some debugging output after the app. 

obrazek

danielhollas (approved):

unkcpz:
I think we didn't have this issue because the single app widget is always not very big. With some output in the cell, the issue will happen. The change is needed.

@danielhollas thanks for checking and approving.

PR #148, opened 2023-03-16
Add detached check before raise imcompatible warning.

unkcpz:
fixes #https://github.com/aiidalab/aiidalab/issues/373

The detached check works fine but since it is detached the version will be `UNKNOWN` and in the compatible check it is regarded as incompatible. I also add a docstring in https://github.com/aiidalab/aiidalab/pull/374

### DIFF
--- a/home/app_manager.py
+++ b/home/app_manager.py
@@ -303,7 +303,12 @@ class AppManagerWidget(ipw.VBox):
             ) and not self.blocked_ignore.value
 
             # Check the compatibility of current installed version and show banner if not compatible.
-            if not busy and installed and not self.app.compatible:
+            if (
+                not busy
+                and installed
+                and not self.app.compatible
+                and not self.app.detached
+            ):
                 self.header_warning.show(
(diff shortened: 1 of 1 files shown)


danielhollas:
@unkcpz I do not think this is the correct approach to fix this. In this way, detached applications would always be compatible, but that's not true in general.

We should instead modify the `is_incompatible` function to work properly. Specifically, when the app is detached, we should pull the dependencies from the `setup.cfg`. We are already doing that somehow when the app is not registered so it should not be a completely new code. Can you take a look at how this all works for unregistered apps? That should give you a hint how to proceed,

unkcpz:
Ah, I think you are correct, AGAIN :shrug: 

The compatibility check is only done for app not registred, we need to extend it to also installed and detached. I'll update the PR later.

unkcpz:
Hi @danielhollas, I did the change in https://github.com/aiidalab/aiidalab/pull/374 and this one is not needed anymore.

PR #154, opened 2023-07-18
Disable prompt for appstore and single app page

danielhollas:
This is a follow up on #141 where we disabled the alert prompt when leaving the home page. We've been using it for quite a while since and haven't noticed any issues so it's time to apply to other apps / pages.

In the future, we should ideally apply this fix directly in Appmode, but that will require more work so for now here's just a simple hotfix.


danielhollas:
@unkcpz by the way, if you haven't already, you might want to apply this fix for the QeApp as well...

yakutovicha (approved):
Tested - seems to work well. Thanks, @danielhollas!

PR #156, opened 2023-07-20
Implement AiiDA control page

### LINKED ISSUE #155: Implement AiiDA-control page
This page should provide user-friendly tools to interact with AiiDA. The features might include:

- Interacting with processes
- Interacting with database
- Interacting with daemon
- Interacting with groups
- ...

yakutovicha:
fixes #155


cpignedoli:
I write here my thoughts on what this app should offer to we worth teh effort wrt verdi process list -a
Let's consider "main-workflow" (MW) a workflow, submitted by a user that usually has a description and that usually submits child workflows.

From a user perspective, after I submitted 2-3 MW (in a day, 20-30 in a month) e,g, "PDOS DBBR on Au", "STM goblet", "DBBR adsorption energy" I would like to, somehow in order of priority:

1. See with a glance on a list that "PDOS..", "STM gob..", ".." MWs are present (focus on the description, not on the aiida workchain name) and in status...
2. I would like to see, e.g. pressing a button on a specific MW,  which steps have been already done, and which one is currently running/queued/paused...
3. I would like to be able to play paused processes, kill...
4. I would like to be able to find the MW "DBBR Br removed on Au" that run 2 months ago, and be able to access the different steps. If I do not remember the name... I would like to be able to see all MWs in a range of dates and be able to scroll them: NO automatic refresh of the page (now in the "process list" app, there is automatic refresh thus, if you scroll a list of 100 processes, quite likely before you can find the one you are looking for the page gets refreshed..)

The app should be "fast"

unkcpz:
The control page dependent on the `aiidalab-widgets-base`, uses widgets from AWB. It is still an open question whether AWB should be the dependency of `aiidalab-home`. 
A problem I can foresee is since app share the environment with all kernel packages, install/update app will potentially break the control page.

We can also make AWB the core package as defined in `aiidalab` and prevent user to override with incompatible version.   But this brings restriction to app developers.

The ultimate solution could be for every app has its own Python environment created when using `aiidalab install` by create an independent conda environment. I had played with this idea a bit but stuck at the jupyter kernel as well as the python interpereter's should be set into notebook metadata in advance which makes no sense for app developement.

danielhollas:
@yakutovicha thanks for this. :clamp: I like the overall approach and design. I would suggest two things so we can move with an initial version of the control page:

1. Vendor in AWB dependencies. This will allow faster iteration on the design, and in the meantime we can figure out whether / how to add AWB to base image.
2. It is clear that the Process page will require the most discussion. Therefore I would omit it in the first version, so that it can be discussed separately and we can have something basic in the meantime. 

Couple minor things / ideas...

1. Perhaps it would make sense to merge the Status and Daemon pages? I would imagine most of the time I'd like to see both at the same time to get the overall state of the system and be able to control it.
2. Provide buttons to increase / decrease number of daemon workers
3. Daemon page: Start / Stop buttons should be deactivated when the daemon is (not) running.
4. Profile page: "Make default" button should not be active for default profile.

danielhollas:
It would be great to re-enable the tests in CI before this PR is merged, per #161, and add basic tests for the new functionality.

control.ipynb
```diff
@@ -0,0 +1,91 @@
...
+    "    value=None,\n",
+    "    orientation=\"vertical\",\n",
+    "    layout=ipw.Layout(width=\"200px\"),\n",
+    ")\n",
+    "\n",
+    "\n",
+    "output = ipw.Output()\n",
+    "\n",
+    "\n",
+    "def update_output(value={\"new\": \"Group\"}):\n",
```
danielhollas, line 56: I've merged in main with updated ruff that now checks jupyter notebooks. It reports a lint violation here:

```

control.ipynb:cell 3:19:25: B006 Do not use mutable data structures for argument defaults
   |
19 | def update_output(value={"new": "Group"}):
   |                         ^^^^^^^^^^^^^^^^ B006
20 |     if value["new"] in correspondance:
21 |         with output:
   |
   = help: Replace with `None`; initialize within function
```

danielhollas:
@yakutovicha btw: I am going to be on the coding retreat next month and I've heard you'll be there as well? Perhaps this PR is a good candidate to work on, WDYT?

danielhollas:
> The control page dependent on the aiidalab-widgets-base, uses widgets from AWB. It is still an open question whether AWB should be the dependency of aiidalab-home.
A problem I can foresee is since app share the environment with all kernel packages, install/update app will potentially break the control page.

Just to continue this discussion, I think `aiidalab-home` cannot depend on AWB. All the apps (including home app) currently share a python environment, which means only one AWB version at a time. But we need to allow application to install different versions of AWB according to their needs, which will always run a risk of breaking home app. Thus I don't see any other way.

If there are components that are currently needed in AWB and needed here, I think the simplest thing is to copy them. I think in general we don't want to expand the home app too much, so the number of such components should not be high.

@edan-bainglass  @superstar54 FYI about the status of this PR, please have a look at my earlier comment. https://github.com/aiidalab/aiidalab-home/pull/156#issuecomment-1667004299

I don't think it would be too much work to finish this PR in some minimal form and keep iterating on it in subsequent PRs.

edan-bainglass:
> @edan-bainglass @superstar54 FYI about the status of this PR, please have a look at my earlier comment. [#156 (comment)](https://github.com/aiidalab/aiidalab-home/pull/156#issuecomment-1667004299)
> 
> I don't think it would be too much work to finish this PR in some minimal form and keep iterating on it in subsequent PRs.

Thanks @danielhollas. @yakutovicha can we set up a meeting soon (maybe before the break) to discuss the current status and define a roadmap to a minimal release? All are welcome to join.

yakutovicha:
> Thanks @danielhollas. @yakutovicha can we set up a meeting soon (maybe before the break) to discuss the current status and define a roadmap to a minimal release? All are welcome to join.

I'm happy to meet, but frankly, the only time that might work for me is this Wed between 16 and 17. All the remaining days are booked.

edan-bainglass:
@yakutovicha this is great! I've tested a few things here and there. Looks good to me. I made a few minor updates in a PR to this PR (#231). Other than that, my only concern is the obvious duplication of the process list:
- QE app provides rich process table
- I see that now we have a process list link at the top of the app
- And now the one introduced here

Ideally, I would lean toward using the richer process table, though in its current state, there are some QE-specific parts. Perhaps this can be left for a later PR if we come to an agreement. As for the one at the top of the app and the new one in the control panel, I would lean toward only the control panel one, making sure that it covers all the features the other provides. 

If we agree to remove the top process list link in favor of the control panel one, I would recommend to elevate the remaining code setup link to a tile button (icon TBD). This of course can be done in a separate PR. Let's get all of this done on our upcoming coding day 🙂

edan-bainglass:
A few other comments that I see now:
- The new process list behaves differently than the one in the QE app in the sense that clicking on the PK takes you to the process notebook, not the QE app loaded with the workflow. This is more general and should be kept this way. Might be a justification for keeping both lists. TBD
- In the daemon log under the daemon panel, we get the process report. However, we also get deprecation warnings. It would be nice to introduce the following (in a follow up PR):
  - symmetry with `verdi process report`, which does not expose deprecation warnings (actually, I'm wondering now if the daemon report covers more than just process reporting?)
  - dropdown selector of report level (default to report, but can be set to debug when the app misbehaves - very useful for assisting users)
  - a copy button (again, helpful when assisting users)
  - a reset/clear button?

yakutovicha:
@edan-bainglass, thanks a lot for the feedback.

To simplify the review process, I will split this PR into smaller ones and will implement each feature one by one. This Particular PR I won't merge. This was really a vibe-coded prototype.

What concerns the process, I agree to keep only the one in the process table. @cpignedoli is eager to fix some functionality of the process list, so we can do that as well.

If you have time, feel free to open the PR to migrate the QE process list here - I checked it, and I like it a lot 👍

Issue #149, opened 2023-03-25
Remove the widgetsnbextension version constraint when issue with `ipytree` is fixed


yakutovicha:
This is to remember that the `widgetsnbextension<3.6.3` constraint was set to avoid [this issue](https://github.com/jupyter-widgets/ipywidgets/issues/3735). Should be removed when things are fixed.

danielhollas:
Based on the discussion in the linked issue, it seems that this was fixed 6 months ago. Would be good to test this.

danielhollas:
Reopening since we needed to revert the pin removal, see aiidalab/aiidalab-docker-stack#512

danielhollas:
> This is to remember that the widgetsnbextension<3.6.3 constraint was set to avoid https://github.com/jupyter-widgets/ipywidgets/issues/3735.

Just noting that looking through the issues, it looks like the fix was released in ipywidgets=7.8.1
https://github.com/jupyter-widgets/ipywidgets/commit/4187fdfbf8bea253e8934622b1a4e583adeb2c18

PR #172, opened 2024-08-12
Fix invalid python requirement

danielhollas:
This was causing an exception in `aiidalab` package, see: 

Alternatively, we might be able to remove this pin altogether, see last comment in #149 

Corresponding fix to be more resilient in `aiidalab` is https://github.com/aiidalab/aiidalab/pull/444


.pre-commit-config.yaml
```diff
@@ -26,11 +26,6 @@ repos:
       hooks:
           - id: yamlfmt
 
-    - repo: https://github.com/asottile/setup-cfg-fmt
```
danielhollas, line 29: We've already removed this hook in other repos, it's more annoying than useful.

.github/workflows/release.yml
```diff
@@ -9,7 +9,7 @@ jobs:
         runs-on: ubuntu-latest
 
         steps:
-            - uses: softprops/action-gh-release@v2.0.6
```
danielhollas, line 12: We've don't pin other other actions to exact versions, and this just creates noise from dependabot.

danielhollas:
Going to land this so I can cut a new release.

PR #176, opened 2024-10-14
Pin notebook version to v6

danielhollas:
We currently don't support Notebook v7 (tracked in #177). Until we do, let's pin the notebook version. (currently, running `pip install .` in empty environment installs notebook v7)


yakutovicha (approved):
👍

PR #178, opened 2024-10-14
Reenable notebook tests

### LINKED ISSUE #161: re-enable tests in CI
There are a couple of basic selenium tests in `tests/` folder, but they are currently not being run in GHA.

These tests need to run inside the aiidalab docker image. We should port the notebook test infrastructure from AWB and re-enable the tests in CI.

danielhollas:
Fixes #161


.github/workflows/ci.yml
```diff
@@ -3,25 +3,66 @@
...
+
+# https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+    # only cancel in-progress jobs or runs for the current workflow - matches against branch & tags
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
 
 jobs:
 
-    pre-commit:
```
danielhollas, line 10: The pre-commit job is not needed since we run it through `pre-commit.ci` app anyway. (we've removed this job in all other repos as well)

tests_notebooks/conftest.py
danielhollas, line 1: Copied almost verbatim from AWB, probably doesn't need to be reviewed

tests_notebooks/conftest.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@pytest.fixture(scope="function")
+def selenium_driver(selenium, notebook_service):
+    def _selenium_driver(nb_path, url_params=None):
+        url, token = notebook_service
+        url_with_token = urljoin(url, f"apps/apps/home/{nb_path}?token={token}")
+        if url_params is not None:
+            for key, value in url_params.items():
+                url_with_token += f"&{key}={value}"
```
danielhollas, line 83: This is new

tests_notebooks/test_manage_app.py
```diff
@@ -0,0 +1,10 @@
+from selenium.webdriver.common.by import By
+
+
+def test_single_app(selenium_driver, final_screenshot):
+    url_params = {"app": "aiidalab-widgets-base"}
+    selenium = selenium_driver("single_app.ipynb", url_params)
+    selenium.set_window_size(1440, 828)
+    selenium.set_window_size(950, 1000)
```
yakutovicha, line 8: Why is it needed to set window size two times?

.github/workflows/ci.yml
yakutovicha: makes sense 👍

yakutovicha (changes requested):
Thanks, @danielhollas!

I have only one question: why do you resize the window two times (see my comments below)? The rest seems to be alright.

tests_notebooks/test_manage_app.py
danielhollas: Whoops, just an ommission on my part, fixed.

yakutovicha (approved):
Go ahead 👍

PR #181, opened 2024-11-19
Fix crash in single_app.ipynb from invalid version

### LINKED ISSUE #180: single_app.ipynb crashes with `packaging>=22.0`
This is the same issue that I recently solved in the `aiidalab` and `aiidalab-registry` packages, see aiidalab/aiidaslab#339 and aiidalab/aiidalab-registry#103 for details. It did not occur to me if the same issues occurs here right in our AiiDAlab home! 

The problem is that the new `packaging` version actually rejects invalid versions, and it just so happens we have one such invalid version for AWB.

```python
---------------------------------------------------------------------------
InvalidVersion                            Traceback (most recent call last)
/tmp/ipykernel_5690/3649478905.py in <cell line: 3>()
      1 app_base = AiidaLabApp(name, app_data, AIIDALAB_APPS)
      2 
----> 3 display(AppManagerWidget(app_base, minimalistic=False))

~/local/aiidalab-home/home/app_manager.py in __init__(self, app, minimalistic)
    208             self._refresh_prereleases, names=["has_prereleases", "installed_version"]
    209         )
--> 210         self._refresh_prereleases(change={"owner": self.app})  # initialize
    211 
    212         children = [

~/local/aiidalab-home/home/app_manager.py in _refresh_prereleases(self, change)
    260         has_prereleases = app.has_prereleases
    261         prerelease_installed = (
--> 262             parse(installed_version).is_prerelease
    263             if isinstance(installed_version, str)
    264             else False

~/.local/lib/python3.9/site-packages/packaging/version.py in parse(version)
     54     :raises InvalidVersion: When the version string is not a valid version.
     55     """
---> 56     return Version(version)
     57 
     58 
~/.local/lib/python3.9/site-packages/packaging/version.py in __init__(self, version)
    200         match = self._regex.search(version)
    201         if not match:
--> 202             raise InvalidVersion(f"Invalid version: {version!r}")
    203 
    204         # Store the parsed out pieces of the version

InvalidVersion: Invalid version: ''
```

danielhollas:
This is the same issue that I recently solved in the aiidalab and aiidalab-registry packages,
see aiidalab/aiidaslab#339 and aiidalab/aiidalab-registry#103 for details.

It did not occur to me that the same issue occurs here right in our AiiDAlab home!

Fixes #180

I'll trigger a rebuild of the Python 3.11 docker image with this branch to test this.

### DIFF
--- a/single_app.ipynb
+++ b/single_app.ipynb
@@ -78,17 +78,7 @@
    "source": [
     "app_base = AiidaLabApp(name, app_data, AIIDALAB_APPS)\n",
     "\n",
-    "try:\n",
-    "    display(AppManagerWidget(app_base, minimalistic=False))\n",
-    "except Exception as error:\n",
-    "    display(\n",
-    "        ipw.HTML(\n",
-    "            '<div style=\"font-size: 30px; text-align:center;\">'\n",
-    "            f\"Unable to show app widget due to error: {error}\"\n",
-    "            \"</div>\",\n",
-    "            layout={\"width\": \"600px\"},\n",
-    "        )\n",
-    "    )"
+    "display(AppManagerWidget(app_base, minimalistic=False))"
    ]
   }
  ],
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ name: continuous-integration
 on:
     push:
         branches:
-            - master
+            - main
     pull_request:
 
 env:
--- a/home/app_manager.py
+++ b/home/app_manager.py
@@ -7,7 +7,7 @@ import traitlets
 from aiidalab.app import AppRemoteUpdateStatus as AppStatus
 from aiidalab.app import AppVersion
 from jinja2 import Template
-from packaging.version import parse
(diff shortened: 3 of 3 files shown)


single_app.ipynb
```diff
@@ -78,17 +78,7 @@
    "source": [
     "app_base = AiidaLabApp(name, app_data, AIIDALAB_APPS)\n",
     "\n",
-    "try:\n",
-    "    display(AppManagerWidget(app_base, minimalistic=False))\n",
-    "except Exception as error:\n",
```
danielhollas, line 83: Catching all exceptions like this is really bad, since it is hard to debug and users cannot provide stack traces in their bug reports.

danielhollas:
@yakutovicha  I triggered a new build of pr 455, so a new version of `ghcr.io/aiidalab/full-stack:pr-455` with a fix will be available in ~20 minutes.

danielhollas:
Ok, I've tested the new image and haven't been able to reproduce the error anymore. :tada:

yakutovicha (dismissed):
it works, thanks 👍

PR #182, opened 2024-11-21
Make home notebook full width

edan-bainglass:
Jupyter adds a `max-width: calc(100% - 14ex)` property settings to the `output_subarea` class, which is everywhere. This PR overwrites it with `max-width: none` to allow container widgets to take up the full width of the notebook.

### DIFF
--- a/start.ipynb
+++ b/start.ipynb
@@ -18,6 +18,25 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {
+    "vscode": {
+     "languageId": "html"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "%%html\n",
+    "\n",
+    "<style>\n",
+    "  .output_subarea {\n",
+    "    max-width: none !important;\n",
+    "  }\n",
+    "</style>\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


start.ipynb
```diff
@@ -15,6 +15,25 @@
     "}"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "vscode": {
+     "languageId": "html"
+    }
```
danielhollas, line 24: I am somewhat surprised that the `nbstripout` pre-commit hook does not remove this. I guess it doesn't matter.

start.ipynb
```diff
@@ -15,6 +15,25 @@
...
+   },
+   "outputs": [],
+   "source": [
+    "%%html\n",
+    "\n",
+    "<style>\n",
+    "  .output_subarea {\n",
+    "    max-width: none !important;\n",
+    "  }\n",
```
danielhollas, line 33: I am going to trust you on this one. :sweat_smile:

danielhollas (approved):
I love this! :heart: 

To illustrate, this is how my homepage look before this PR:
image

And after:
image

I do think that it would be very valuable to implement this in [appmode](https://github.com/oschuett/appmode) directly. Let's discuss tomorrow if there's time.

start.ipynb
danielhollas: Okay, I guess this is not that complicated, you're just ensuring that the `max-width` parameter is not set.

danielhollas:
Actually, I do have one question here:
> Jupyter adds a max-width: calc(100% - 14ex) property settings to the output_subarea class, which is everywhere. 

What do you mean by "everywhere"?

PR #183, opened 2024-11-21
Add auto-detected home warning message mechanism

### LINKED ISSUE #184: Add a mechanism for custom message in home app (for demo server)
Add demo message (message saying that we are in a demo, how long data is kept, …) in AiiDAlab home (initial page with ‘File manager’ ‘Terminal’ … QEapp logo [Jason and Edan] [copied above]
Make sure the default guide/tutorial is on, in the demo server. [Edan Jason Xing]

edan-bainglass:
Based on #182

This PR adds a mechanism to sense if a `home_app_warning.md` file has been placed in the container's `/home/<user>/.aiidalab` directory, read it if so, and display it between the controls and the apps section.

Closes #184

### DIFF
--- a/home/start_page.py
+++ b/home/start_page.py
@@ -3,6 +3,7 @@
 import json
 from glob import glob
 from os import path
+from pathlib import Path
 
 import ipywidgets as ipw
 import traitlets
@@ -29,7 +30,7 @@ def create_app_widget_move_buttons(name):
 
 
 class AiidaLabHome:
-    """Class that mananges the appearance of the AiiDAlab home page."""
+    """Class that manages the appearance of the AiiDAlab home page."""
 
     def __init__(self):
         self.config_fn = ".launcher.json"
@@ -40,15 +41,15 @@ class AiidaLabHome:
         """Create the widget representing the app on the home screen."""
         config = self.read_config()
         app = AiidaLabApp(name, None, AIIDALAB_APPS)
-
-        if name == "home":
-            app_widget = AppWidget(app, allow_move=False, allow_manage=False)
-        else:
-            app_widget = CollapsableAppWidget(app, allow_move=True)
-            app_widget.hidden = name in config["hidden"]
-            app_widget.observe(self._on_app_widget_change_hidden, names=["hidden"])
-
+        app_widget = CollapsableAppWidget(app, allow_move=True)
+        app_widget.hidden = name in config["hidden"]
+        app_widget.observe(self._on_app_widget_change_hidden, names=["hidden"])
         return app_widget
+
+    def _create_home_widget(self):
+        """Create the home app widget."""
+        app = AiidaLabApp("home", None, AIIDALAB_APPS)
+        return AppWidget(app, allow_move=False, allow_manage=False)
 
     def _on_app_widget_change_hidden(self, change):
         """Record whether a app widget is hidden on the home screen in the config file."""
@@ -72,7 +73,17 @@ class AiidaLabHome:
     def render(self):
         """Rendering all apps."""
 
-        displayed_apps = []
+        home = self._create_home_widget()
+        children = [home]
+
+        config_dir = Path.home() / ".aiidalab"
+        warning_file = config_dir / "home_app_warning.md"
+
+        if warning_file.exists():
+            content = warning_file.read_text()
+            notification = self._create_notification(content)
+            children.append(notification)
+
         apps = self.load_apps()
 
         for name in apps:
@@ -80,8 +91,10 @@ class AiidaLabHome:
             if name not in self._app_widgets:
                 self._app_widgets[name] = self._create_app_widget(name)
 
-            displayed_apps.append(self._app_widgets[name])
-        self.output.children = displayed_apps
+            children.append(self._app_widgets[name])
+
+        self.output.children = children
+
         return self.output
 
     def load_apps(self):
@@ -98,7 +111,7 @@ class AiidaLabHome:
         apps.sort(key=lambda x: order.index(x) if x in order else -1)
         config["order"] = apps
         self.write_config(config)
-        return ["home", *apps]
+        return apps
 
     def move_updown(self, name, delta):
         """Move the app up/down on the start page."""
@@ -111,6 +124,19 @@ class AiidaLabHome:
         config["order"] = order
         self.write_config(config)
 
+    def _create_notification(self, content):
+        from IPython.display import Markdown, display
+        from jinja2 import Environment
+
+        env = Environment()
+        notification = env.from_string(content).render()
+        output = ipw.Output()
+        notification_widget = ipw.VBox(children=[output])
+        notification_widget.add_class("home-notification")
+        with output:
+            display(Markdown(notification))
+        return notification_widget
+
 
 class AppWidget(ipw.VBox):
     """Widget that represents an app as part of the home page."""
@@ -119,7 +145,7 @@ class AppWidget(ipw.VBox):
         self.app = app
 
         launcher = load_widget(app.name)
-        launcher.layout = ipw.Layout(width="900px")
+        launcher.layout.flex = "1"  # fill available space
 
         header_items = []
         footer_items = []
@@ -128,7 +154,7 @@ class AppWidget(ipw.VBox):
             app_status_info = AppStatusInfoWidget()
             for trait in ("detached", "compatible", "remote_update_status"):
                 ipw.dlink((app, trait), (app_status_info, trait))
-            app_status_info.layout.margin = "0px 0px 0px 800px"
+            app_status_info.layout.margin = "0px 0px 0px auto"
             header_items.append(app_status_info)
 
             footer_items.append(
@@ -150,7 +176,7 @@ class AppWidget(ipw.VBox):
 
         footer = ipw.HTML(" ".join(footer_items), layout={"width": "initial"})
         footer.layout.margin = (
-            "0px 0px 0px 700px" if allow_manage else "0px 0px 20px 0px"
+            "0px 0px 0px auto" if allow_manage else "0px 0px 20px 0px"
         )
 
         super().__init__(children=[header, body, footer])
--- a/tests_notebooks/conftest.py
+++ b/tests_notebooks/conftest.py
@@ -56,8 +56,11 @@ def aiidalab_exec(docker_compose):
 
 @pytest.fixture
 def create_warning_file(aiidalab_exec):
+    aiidalab_exec("mkdir -p /home/jovyan/.aiidalab", user="root")
+    aiidalab_exec("chmod a+xr /home/jovyan/.aiidalab", user="root")
     aiidalab_exec(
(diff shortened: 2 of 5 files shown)


danielhollas:
> This PR adds a mechanism to sense if a home_app_warning.md file has been placed in the container's /home/<user>/.aiidalab directory, read it if so, and display it between the controls and the apps section.

Can you explain a bit what is the use case of this? (ideally in an Github issue if it doesn't exist already)

edan-bainglass:
> Can you explain a bit what is the use case of this? (ideally in an Github issue if it doesn't exist already)

GitHub issue was attached to aiidalab-qe repo, now migrated here and mentioned in description 🙂

home/start_page.py
```diff
@@ -56,27 +56,26 @@ def _inner(self, change):
...
-            app_widget = CollapsableAppWidget(app, allow_move=True)
-            app_widget.hidden = name in config["hidden"]
-            app_widget.observe(self._on_app_widget_change_hidden, names=["hidden"])
-
+        app_widget = CollapsableAppWidget(app, allow_move=True)
+        app_widget.hidden = name in config["hidden"]
+        app_widget.observe(self._on_app_widget_change_hidden, names=["hidden"])
         return app_widget
 
+    def _create_home_widget(self):
```
danielhollas, line 74: Nice cleanup, thank you!

home/start_page.py
```diff
@@ -98,18 +97,28 @@ def read_config(self):
...
     def render(self):
         """Rendering all apps."""
-        self.output.clear_output()
-        apps = self.load_apps()
+        home = self._create_home_widget()
+        display(home)
+
+        from pathlib import Path
+
+        config_dir = Path("../../.aiidalab").resolve().absolute()
```
danielhollas, line 105: This feels more robust and simple?

```suggestion
        config_dir = Path.home()  / ".aiidalab"
```

home/start_page.py
```diff
@@ -98,18 +97,28 @@ def read_config(self):
 
     def render(self):
         """Rendering all apps."""
-        self.output.clear_output()
-        apps = self.load_apps()
+        home = self._create_home_widget()
+        display(home)
+
+        from pathlib import Path
```
danielhollas, line 103: Let's.move this import at the top of the module, not much point in inlining this imo. If you disagree, please at least move at the top of the function.

home/start_page.py
danielhollas: Also this should probably be a helper function somewhere outside of this notebook. Is this directory used anywhere else in home app?

danielhollas (changes requested):
Code looks great, thank you for the cleanup as well. I do think this needs a  test. I have just recently re-enabled selenium tests in this repo, using the same fixtures as in AWB, ao hopefully shouldn't be too much work.

danielhollas:
This needs to be updated with changes on main.

> I do think this needs a test. I have just recently re-enabled selenium tests in this repo, using the same fixtures as in AWB, ao hopefully shouldn't be too much work.

Let's not block this PR on tests in case writing one turns out to be more work, we can do it in a follow-up. Once this is merged I'll push a new aiidalab-home version.

home/start_page.py
```diff
@@ -146,7 +168,7 @@ def __init__(self, app, allow_move=False, allow_manage=True):
         self.app = app
 
         launcher = load_widget(app.name)
-        launcher.layout = ipw.Layout(width="900px")
+        launcher.layout.flex = "1"
```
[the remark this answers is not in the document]
edan-bainglass: `flex` is shorthand for three flex-related properties. If you give a single value, it goes to the `flex-grow` property, which is described [here](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow#description). This is quite technical though. I generally understand it as _assign the remaining free space to this flex item_.

For example, if you have a row with two items, and the row display is set to `flex`, the two items become flex items. If you set the `flex` property of the first item to `"1"`, you are saying "give the remaining free space of the row to this item. So if the two items together only take 20% of the row width (let's say 10% each), setting `flex` to `"1"` on the first will yield a 90/10 split, as the remaining 80% are allocated to the first item.

Hopefully it is clear that all `flex: 1` means on an item is "please consume the remaining free space".

@danielhollas any clearer?

home/start_page.py
edan-bainglass: Nice! Didn't know about `.home()`. 100% more robust/simple.

As for the `.aiidalab` dir being used elsewhere, maybe not at the moment, but I already have some ideas for other uses. So yes, good to make it easily accessible by some means.

home/start_page.py
edan-bainglass: I can't recall a good reason why it wasn't placed at the module's head. Will move up.

edan-bainglass:
@danielhollas FYI, in and out of vacation this week. But planning to wrap this up by end of week regardless.

home/start_page.py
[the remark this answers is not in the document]
edan-bainglass: Also added an in-line comment 🙂

home/start_page.py
edan-bainglass: Can't think of where to place it. Let's discuss maybe for a separate PR.

edan-bainglass:
@danielhollas can you take a look at my selenium test? Not sure why it can't find the element.

tests_notebooks/test_notification.py
```diff
@@ -0,0 +1,37 @@
...
+from selenium.webdriver.remote.webdriver import WebDriver
+
+
+def test_home_notification(selenium_driver, final_screenshot):
+    config_dir = Path.home() / ".aiidalab"
+    warning_file = config_dir / "home_app_warning.md"
+
+    # If the file exists, we are testing locally
+    # We should make sure not to overwrite or remove the file
+    testing_locally = warning_file.exists()
```
danielhollas, line 14: I think this case would be better handled as a fixture. The test should not even touch / see the normal config dir. Instead you could e.g. have the config dir specified as a global variable, that would get overridden in the fixture with `tmp_path / ".aiidalab"`.

edan-bainglass: Not sure what you mean, but also, maybe you can help me first figure out why the test is not working as is? 😅🙏

danielhollas:
@edan-bainglass when debugging these tests, the first thing is to look at the screenshots that are uploaded as artifacts on GHA, e.g. [here](https://github.com/aiidalab/aiidalab-home/actions/runs/12155024832?pr=183)

Looking at this particular test, I can see there is no warning rendered
 
image

LMK in case I can be more helpful (I don't see anything obviously wrong in the test)

edan-bainglass:
> @edan-bainglass when debugging these tests, the first thing is to look at the screenshots that are uploaded as artifacts on GHA, e.g. [here](https://github.com/aiidalab/aiidalab-home/actions/runs/12155024832?pr=183)
> 
> Looking at this particular test, I can see there is no warning rendered
> 
> LMK in case I can be more helpful (I don't see anything obviously wrong in the test)

Right, I always forget about the screenshot. But yes, the issue is why is it not rendering the warning widget? I added an assert after creating the markdown file to ensure it is indeed there, which passes. So the file is there and the mechanism to render the file is there. Works locally. Not sure why it isn't passing.

I don't have much experience with selenium tests. It is unclear to me why something that works locally does not work remotely. Is there some refresh I need to trigger? Docs aren't clear. I think we should have our own docs for writing selenium, considering that it is often required for PRs.

danielhollas:
> I added an assert after creating the markdown file to ensure it is indeed there, which passes. So the file is there and the mechanism to render the file is there

Oh, that got me thinking, I don't think you're placing the file in the correct place. I suspect if you try to run the test locally, you'll see the file created in your local folder, but it needs to be created inside the container. 

I think you need to use the `aiidalab_exec` fixture to create the file inside the container. Something like this

```python
@pytest.fixture
def create_warning_file(aiidalab_exec):
      aiidalab_exec("echo 'This is a test warning' > /home/jovyan/.aiidalab/home_app_warning.md")
```

edan-bainglass:
> > I added an assert after creating the markdown file to ensure it is indeed there, which passes. So the file is there and the mechanism to render the file is there
> 
> Oh, that got me thinking, I don't think you're placing the file in the correct place. I suspect if you try to run the test locally, you'll see the file created in your local folder, but it needs to be created inside the container.
> 
> I think you need to use the `aiidalab_exec` fixture to create the file inside the container. Something like this
> 
> ```python
> @pytest.fixture
> def create_warning_file(aiidalab_exec):
>       aiidalab_exec("echo 'This is a test warning' > /home/jovyan/.aiidalab/home_app_warning.md")
> ```

Still raising an issue. I think this closer to what we should do, but I guess I'm still not doing it correctly. If you understand how our fixtures work, can you please resolve this matter? And can we please document how to use our testing fixtures? Especially when it comes to selenium?

danielhollas:
@edan-bainglass is it okay if I push to your branch? I can try something.

I can hear your frustration, and these tests are a pain. I do agree we need to document things better. One thing I'd like to do is to move from Selenium to Playwright (https://github.com/aiidalab/aiidalab-widgets-base/issues/524) which should work better, has better docs and should have much better support for local testing. But in this case you're running more into issues with our Docker setup than anything else.

Did you try to run the tests locally? I believe if you have docker-compose installed it should work. It might need some extra installation for browser support. @yakutovicha I believe you were able to run selenium tests locally, would you mind to chime in here?

edan-bainglass:
@danielhollas go for it 👍

tests_notebooks/conftest.py
```diff
@@ -41,25 +41,39 @@ def docker_compose(docker_services):
...
-            command = f"exec --workdir {workdir} -T aiidalab {command}"
+def aiidalab_exec(docker_compose, nb_user):
+    """Execute command inside the AiiDAlab test container"""
 
+    def execute(command, user=None, **kwargs):
+        workdir = f"/home/{nb_user}/apps/home"
+        if user is None:
+            user = nb_user
+        command = (
+            f"exec --workdir {workdir} -T --user={user} aiidalab bash -c '{command}'"
```
danielhollas, line 52: This is the important fix, we need to escape the commands passed to docker compose, using `bash -c '<command>'`

danielhollas (approved):

danielhollas:
The tests are passing! :tada: 

image

The issue was annoying :anguished: Basically, in `aiidalab_exec` fixture we were passing the command directly to `docker-compose`, but because we were not escaping the command in any way, it would not work for more complex commands that include for example redirection chaining commands with `&&`. The solution to this is to wrap the command in `bash -c 'command'`.

You can easily see to problem by running the following locally

```
❯ docker run aiidalab/full-stack ls && cd apps/
apps
work
bash: cd: apps/: No such file or directory
```

versus

```
❯ docker run aiidalab/full-stack bash -c 'ls && cd apps/'
apps
work
```

@edan-bainglass have a look. All good from my side.

edan-bainglass:
@danielhollas wow! What a struggle. Glad you were able to figure it out. Maybe we should have a chat soon to discuss a solid strategy for AiiDAlab app testing, even if it just means solid docs.

Anyhow, thanks for the work. Looks good!

PR #185, opened 2024-11-22
Start page: use `ipw.VBox` instead of `ipw.Output` to display apps.

yakutovicha:
This also allows to remove hotfix to observe collapsing apps.

### DIFF
--- a/home/start_page.py
+++ b/home/start_page.py
@@ -1,7 +1,6 @@
 """Module to generate AiiDAlab home page."""
 
 import json
-from functools import wraps
 from glob import glob
 from os import path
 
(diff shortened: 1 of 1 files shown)


home/start_page.py
```diff
@@ -186,24 +185,17 @@ def __init__(self, app, allow_move=False, allow_manage=True):
...
         self.app = app
         app_widget = AppWidget(app, **kwargs)
         super().__init__(children=[app_widget])
         self.set_title(0, app.title)
-        # Need to observe all names here due to unidentified issue:
-        self.observe(
-            self._observe_accordion_selected_index
-        )  # , names=['selected_index'])
 
-    @_workaround_property_lock_issue
```
danielhollas, line 201: Great! Please remove this decorator (defined in this file) since it is not used anywhere else.

danielhollas (changes requested):
Looks great, thank you for this cleanup @yakutovicha! My only request is to delete the now unused workaround decorator.

I think this change conflicts with #184 since it touches the same code, I think it would be good to merge this one first, since it might simplify the implementation in #184. CC @edan-bainglass d

danielhollas (approved):

edan-bainglass:
@yakutovicha very nice! @danielhollas thanks for the ping. I'll rebase and adjust/clean up.

PR #186, opened 2024-11-28
Bump ruff to 0.8.3

danielhollas:


tests_notebooks/conftest.py
```diff
@@ -6,7 +6,7 @@
 import pytest
 import requests
 import selenium.webdriver.support.expected_conditions as ec
-from requests.exceptions import ConnectionError
+from requests.exceptions import ConnectionError  # noqa: A004
```
danielhollas, line 9: ruff was complaining that `ConnectionError` shadows a builtin exception.

PR #189, opened 2024-12-14
Replace `jinja2` with `markdown` for loading the warning message document

edan-bainglass:
This PR fixes an instability with using `jinja2` and `IPython.display.Markdown` to load the warning message document. Here we instead use `markdown.Markdown` to convert the markdown document to HTML and render it using an `ipywidget.HTML` widget, which is simpler to work with (no `ipywidgets.Output` widgets required!).

When I add the `.aiidalab/home_app_warning.md` file, instead of adding the message after the home's control panel, it replaces the panel.

Before adding file
image

After adding file
image

In this PR:

image

### DIFF
--- a/home/start_page.py
+++ b/home/start_page.py
@@ -125,16 +125,12 @@ class AiidaLabHome:
         self.write_config(config)
 
     def _create_notification(self, content):
-        from IPython.display import Markdown, display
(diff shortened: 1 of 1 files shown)


danielhollas (approved):
LGTM, thanks!

PR #197, opened 2025-01-31
Update app container styles

edan-bainglass:


mikibonacci (approved):
Hi @edan-bainglass, LGTM!

danielhollas:
@edan-bainglass just a small ask for a future, for the changes that affect UI, can you please always include before / after screenshots? Would be much easier for reviewers to have a look without needing to check out the changes locally. (e.g. here it's not clear what changed / what was the motivation)

PR #198, opened 2025-03-04
Implement and utilize countdown widget (for temporary deployments)

edan-bainglass:
This PR implement a count down widget for use in temporary deployments. The widget is added to the main notebook within a check for a demo-server-config.yml file.

To test, add `~/.aiidalab/demo-server-config.yml` with `duration: "##:##:##"` in it. Try different times. In particular, try near 5 minutes to check the color change, and near 0 to see the final message.

Checking elapsed time is done via `ps -o etime= -p 1`, as suggested by @yakutovicha 

---

image

image

image

### DIFF
--- a/home/widgets.py
+++ b/home/widgets.py
@@ -197,8 +197,8 @@ class CountDownWidget(ipw.VBox):
     TEMPLATE = "<h1 style='color: {color}'>Time remaining: <b>{countdown}</b></h1>"
     FINAL_WARNING = "<h1 style='color: red'>AiiDAlab container shutdown imminent - please save your work!</h1>"
 
-    def __init__(self, duration: str, **kwargs):
-        self.reference_time = datetime.strptime(duration, "%H:%M:%S")
+    def __init__(self, lifetime: str, **kwargs):
+        self.reference_time = datetime.strptime(lifetime, "%H:%M:%S")
 
         self.countdown = ipw.HTML()
 
--- a/home/widgets.py
+++ b/home/widgets.py
@@ -1,7 +1,7 @@
 """AiiDAlab basic widgets."""
 
-from datetime import datetime
 import subprocess
+from datetime import datetime
 from threading import Timer
 
 import ipywidgets as ipw
--- a/home/widgets.py
+++ b/home/widgets.py
@@ -1,5 +1,7 @@
 """AiiDAlab basic widgets."""
 
+from datetime import datetime
+import subprocess
 from threading import Timer
 
 import ipywidgets as ipw
@@ -187,3 +189,53 @@ class LogOutputWidget(ipw.VBox):
             self._output.value = (
                 self.template.format(text=self.value) if self.value else ""
             )
+
+
+class CountDownWidget(ipw.VBox):
+    """A countdown w.r.t the AiiDAlab container creation time."""
+
+    TEMPLATE = "<h1 style='color: {color}'>Time remaining: <b>{countdown}</b></h1>"
+    FINAL_WARNING = "<h1 style='color: red'>AiiDAlab container shutdown imminent - please save your work!</h1>"
+
+    def __init__(self, duration: str, **kwargs):
+        self.reference_time = datetime.strptime(duration, "%H:%M:%S")
+
+        self.countdown = ipw.HTML()
+
+        super().__init__(
+            children=[
+                self.countdown,
+            ],
+            **kwargs,
+        )
+
+    def start(self):
+        Timer(1, self._update_countdown).start()
+
+    def _update_countdown(self):
+        try:
+            remaining, time_running_out = self._remaining_time()
+            message = self.TEMPLATE.format(
+                color="red" if time_running_out else "black",
+                countdown=remaining,
+            )
+        except Exception:
+            remaining = None
+            message = "Failed to obtain remaining time."
+        self.countdown.value = message
+        if remaining != "00:00:00":
+            self.start()
+        else:
+            self.countdown.value = self.FINAL_WARNING
+
+    def _remaining_time(self) -> tuple[str, bool]:
+        process = subprocess.check_output(["ps", "-o", "etime=", "-p", "1"])
+        elapsed_str = process.decode().strip()
+        if len(elapsed_str) < 6:
+            elapsed_str = f"00:{elapsed_str}"
+        elapsed_time = datetime.strptime(elapsed_str, "%H:%M:%S")
+        remaining = self.reference_time - elapsed_time
+        if remaining.total_seconds() >= 0:
+            time_running_out = remaining.total_seconds() < 60 * 5
+            return (datetime.min + remaining).strftime("%H:%M:%S"), time_running_out
+        return "00:00:00", True
--- a/start.ipynb
+++ b/start.ipynb
@@ -52,6 +52,26 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "from yaml import safe_load\n",
+    "\n",
+    "from home.widgets import CountDownWidget\n",
+    "\n",
+    "# For temporary deployments (e.g. demo server), duration is read from a local\n",
+    "# config file. This will initiate the countdown widget.\n",
+    "demo_server_config_file = Path.home() / \".aiidalab\" / \"demo-server-config.yml\"\n",
+    "if demo_server_config_file.exists():\n",
+    "    demo_server_config = safe_load(demo_server_config_file.read_text())\n",
+    "    countdown = CountDownWidget(duration=demo_server_config.get(\"duration\", \"12:00:00\"))\n",
+    "    countdown.start()\n",
+    "    display(countdown)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "from home.start_page import AiidaLabHome\n",
     "\n",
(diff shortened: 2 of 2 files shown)


edan-bainglass:
@superstar54 @mikibonacci can I get a review please? 🙏 Pinging @danielhollas also if he has time 🙂

start.ipynb
```diff
@@ -46,6 +46,26 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from yaml import safe_load\n",
+    "\n",
+    "from home.widgets import CountDownWidget\n",
+    "\n",
+    "# For temporary deployments (e.g. demo server), duration is read from a local\n",
+    "# config file. This will initiate the countdown widget.\n",
+    "demo_server_config_file = Path.home() / \".aiidalab\" / \"demo-server-config.yml\"\n",
```
superstar54, line 61: Similar to the QEapp, we don't want to hardcode the `demo-server` in the app, so I would suggest using `aiidalab-home.yaml`, or `aiidalab.yaml` directly.

start.ipynb
```diff
@@ -46,6 +46,26 @@
...
+    "if demo_server_config_file.exists():\n",
+    "    demo_server_config = safe_load(demo_server_config_file.read_text())\n",
+    "    countdown = CountDownWidget(duration=demo_server_config.get(\"duration\", \"12:00:00\"))\n",
```
superstar54, line 64: - change `duration` to a better name, e.g., `container_lifetime`
- the lifetime does not set explicitly in the config file, the default value should be a very large time (e.g., 10 years), instead of `12:00:00`.

home/widgets.py
```diff
@@ -187,3 +189,53 @@ def _refresh_output(self, _=None):
...
+            self.start()
+        else:
+            self.countdown.value = self.FINAL_WARNING
+
+    def _remaining_time(self) -> tuple[str, bool]:
+        process = subprocess.check_output(["ps", "-o", "etime=", "-p", "1"])
+        elapsed_str = process.decode().strip()
+        if len(elapsed_str) < 6:
+            elapsed_str = f"00:{elapsed_str}"
+        elapsed_time = datetime.strptime(elapsed_str, "%H:%M:%S")
```
superstar54, line 236: Also consider time longer than one day, (maybe even months).

superstar54 (changes requested):
Hi @edan-bainglass , I missed the previous review request since I get flooded with hundreds of GitHub (action) emails daily. :rofl: .

See my comments below.

home/widgets.py
edan-bainglass: What use case are you thinking of here? I'm not trying to make this general. This is specifically for demo server deployments.

superstar54: I understand that our current demo server only supports 12 hours, but since this is in the AiiDAlab-home, it should be more general, because other people may use this feature, or, we may use this feature in another deployment longer than 12 hours.

edan-bainglass: We can update this for "other people" down the road. For now, this is fine for our use case.

start.ipynb
edan-bainglass: Actually, if the duration is not set in the config file, there should be no countdown widget deployed. I will adjust the conditional.

start.ipynb
edan-bainglass: Actually, this is explicitly for demo servers (not our demo server, but for demo servers)

home/widgets.py
superstar54: Just more caution for the aiidalab-home because every time we make a new release, the docker image must be upgraded. 

But OK to me if you intend to leave this for the future.

danielhollas:
@edan-bainglass can you provide screenshots of how this looks? Is the countdown displayed only in the home page?

edan-bainglass:
> @edan-bainglass can you provide screenshots of how this looks? Is the countdown displayed only in the home page?

Updated description with images. And yes, only on the home page.

edan-bainglass:
> > @edan-bainglass can you provide screenshots of how this looks? Is the countdown displayed only in the home page?
> 
> Updated description with images. And yes, only on the home page.

Hi @danielhollas. Did you have a question or comment regarding this PR?

danielhollas:
@edan-bainglass appologies for late answer, I've been moving to a new flat today so last week has been busy. :-) 

>  What use case are you thinking of here? I'm not trying to make this general. This is specifically for demo server deployments.

My main concern here is that this doesn't feel like something that should live in `aiidalab-home`, at least not in its current form, as it is very specific to the upcoming demo server. I think it would be much better if this could somehow be part of QeApp image, which would allow you to iterate quickly regardless of aiidalab-home and full-stack image.

For example, you could have this as part of the QeApp's `start.py`.

edan-bainglass:
> @edan-bainglass appologies for late answer, I've been moving to a new flat today so last week has been busy. :-)
> 
> > What use case are you thinking of here? I'm not trying to make this general. This is specifically for demo server deployments.
> 
> My main concern here is that this doesn't feel like something that should live in `aiidalab-home`, at least not in its current form, as it is very specific to the upcoming demo server. I think it would be much better if this could somehow be part of QeApp image, which would allow you to iterate quickly regardless of aiidalab-home and full-stack image.
> 
> For example, you could have this as part of the QeApp's `start.py`.

This would limit the countdown to the QE App home app container. If for some reason the demo server user uninstalls the QE App (unlikely), no more countdown.

I understand that the countdown widget needs work to make it available for more use cases. However, I fail to see how this is an issue at the present stage. There have not been any use cases other than the demo server, so no hypothetical use cases will miss the fact that it is not yet general. Placing it here will immediately make it useful to its first use case. When time permits, we iterate, generalize, and release a patch or minor release. Please explain the practical reasons why you believe this shouldn't be accepted. I'm happy to change my opinion if I'm overlooking something.

edan-bainglass:
> > @edan-bainglass appologies for late answer, I've been moving to a new flat today so last week has been busy. :-)
> > > What use case are you thinking of here? I'm not trying to make this general. This is specifically for demo server deployments.
> > 
> > 
> > My main concern here is that this doesn't feel like something that should live in `aiidalab-home`, at least not in its current form, as it is very specific to the upcoming demo server. I think it would be much better if this could somehow be part of QeApp image, which would allow you to iterate quickly regardless of aiidalab-home and full-stack image.
> > For example, you could have this as part of the QeApp's `start.py`.
> 
> This would limit the countdown to the QE App home app container. If for some reason the demo server user uninstalls the QE App (unlikely), no more countdown!
> 
> I understand that the countdown widget needs work to make it available for more use cases. However, I fail to see how this is an issue at the present stage. There have not been any use cases other than the demo server, so no hypothetical use cases will miss the fact that it is not yet general. Placing it here will immediately make it useful to its first use case. When time permits, we iterate, generalize, and release a patch or minor release. Please explain the practical reasons why you believe this shouldn't be accepted. I'm happy to change my opinion if I'm overlooking something.

While you're at it, good if you can bullet point the missing features required in your opinion for this to be accepted here. I can then better assess if it is doable in a reasonable time frame 🙂

danielhollas:
> While you're at it, good if you can bullet point the missing features required in your opinion for this to be accepted here. I can then better assess if it is doable in a reasonable time frame 🙂

I think the main thing is that I would put this in the top header (perhaps next to the AiiDAlab app logo) since then it would be always displayed. If the user open QeApp, then it's going to be very easy to miss the countdown.

I also think that the timer should be implemented on the JS side, the only thing that it needs from the backend is the start time, right?

Right now you're executing subprocess call every second in a background thread, and change the HTML output every second, if I understand correctly. That's a lot of unnecessary traffic. More importantly, this can lead to a severe memory leak if the user closes the browser window unexpectedly (such putting his computer to sleep). We've had huge issues with this in the past, see:
 
https://github.com/aiidalab/issues/issues/13#issuecomment-1523233470

home/widgets.py
```diff
@@ -187,3 +189,53 @@ def _refresh_output(self, _=None):
...
+
+        super().__init__(
+            children=[
+                self.countdown,
+            ],
+            **kwargs,
+        )
+
+    def start(self):
+        Timer(1, self._update_countdown).start()
```
danielhollas, line 213: Infinitely running background threads that update the UI state are unfortunately a really bad idea due to a possible memory leak described in https://github.com/aiidalab/issues/issues/13. 

I am afraid timer like this needs to be in javascript. If I am not mistaken, as long as the timer is initialized properly, it shouldn't need any information from the backend, right?

danielhollas (changes requested):
Disregarding the philosophical issues discussed above, I am afraid this has a fundamental flaw of running infinite background thread that can cause memory leaks. Please have a look at https://github.com/aiidalab/issues/issues/13, I am happy to explain more.

edan-bainglass:
Moving the implementation to a separate app ([aiidalab-timer](https://github.com/edan-bainglass/aiidalab-timer)). @danielhollas @superstar54 please give it a try and comment in this [issue](https://github.com/edan-bainglass/aiidalab-timer/issues/1) 🙏

unkcpz:
Sorry to chime in. I believe there is a way that you can inject javascript and have this javascript override the header by inject the timer I believe? Since we got all permission for the deployment, it should not be a security problem. But I am not expert on js stuff, so I might be totally wrong..

edan-bainglass:
Also, I'm closing this PR. I agree that it should not live in the home app.

PR #207, opened 2025-11-10
Enable ipywidgets 8.x

### LINKED ISSUE #149: Remove the widgetsnbextension version constraint when issue with `ipytree` is fixed

yakutovicha:
fix #149

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,7 @@ install_requires =
     Markdown>=3.4
     aiidalab>=v21.10.2
     ipython~=7.0
-    ipywidgets>=7.8.1,<9
+    ipywidgets~=8.0
     traitlets~=5.0
     notebook~=6.0
 python_requires = >=3.9


setup.cfg
```diff
@@ -18,18 +18,15 @@ classifiers =
...
-# https://github.com/aiidalab/aiidalab-home/issues/149
 [options]
 packages = find:
 install_requires =
     Jinja2>=2.11.3,<4
     Markdown>=3.4
     aiidalab>=v21.10.2
     ipython~=7.0
-    ipywidgets~=7.6
+    ipywidgets>=7.8.1,<9
```
edan-bainglass: So we still support 7.8.1? How come not just >=8?

PR #220, opened 2026-03-16
Consolidate AWB preview handling

yakutovicha:
AWB is still called in some places, but only to display widgets. In this PR we make sure that AWB is only called when it is installed. Otherwise, nothing is shown and a helper message is given to the user.

### DIFF
--- a/home/node_preview.py
+++ b/home/node_preview.py
@@ -1,0 +1,31 @@
+"""Helpers for optional node previews."""
+
+from __future__ import annotations
+
+from aiida import orm
+
+AWB_UNAVAILABLE_MESSAGE = (
+    "The AiiDAlab widgets are not available. Please install it with "
+    "`pip install aiidalab-widgets-base` to use this feature."
+)
+
+
+def _load_awb_viewer():
+    from aiidalab_widgets_base import viewer  # noqa: PLC0415
+
+    return viewer
+
+
+def render_unavailable_preview() -> str:
+    """Return the fallback message shown when AWB is unavailable."""
+    return AWB_UNAVAILABLE_MESSAGE
+
+
+def render_node_preview(node: orm.Node) -> object:
+    """Render a node preview using AWB when it is available."""
+    try:
+        viewer = _load_awb_viewer()
+    except ImportError:
+        return render_unavailable_preview()
+
+    return viewer(node)
--- a/home/node_preview.py
+++ b/home/node_preview.py
@@ -6,8 +6,8 @@ import ipywidgets as ipw
 from aiida import orm
 
 AWB_UNAVAILABLE_MESSAGE = (
-    "The AiiDAlab widgets are not available. Please install it with "
-    "<code>pip install aiidalab-widgets-base</code> to use this feature."
+    "For a richer output view, please install the AiiDAlab widgets with "
+    "<code>pip install aiidalab-widgets-base</code>."
 )
 
 
--- a/home/node_preview.py
+++ b/home/node_preview.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import ipywidgets as ipw
 from aiida import orm
 
 AWB_UNAVAILABLE_MESSAGE = (
@@ -15,6 +16,7 @@ def render_node_preview(node: orm.Node) -> object:
     try:
         from aiidalab_widgets_base import viewer  # noqa: PLC0415
     except ImportError:
-        return AWB_UNAVAILABLE_MESSAGE
-
+        return ipw.HTML(
+            value=str(node) + "<br><em>" + AWB_UNAVAILABLE_MESSAGE + "</em>"
+        )
     return viewer(node)
--- a/home/node_preview.py
+++ b/home/node_preview.py
@@ -7,7 +7,7 @@ from aiida import orm
 
 AWB_UNAVAILABLE_MESSAGE = (
     "The AiiDAlab widgets are not available. Please install it with "
-    "`pip install aiidalab-widgets-base` to use this feature."
+    "<code>pip install aiidalab-widgets-base</code> to use this feature."
 )
 
 
--- a/home/node_preview.py
+++ b/home/node_preview.py
@@ -11,7 +11,7 @@ AWB_UNAVAILABLE_MESSAGE = (
 )
 
 
-def render_node_preview(node: orm.Node) -> object:
+def render_node_preview(node):
     """Render a node preview using AWB when it is available."""
     try:
         from aiidalab_widgets_base import viewer  # noqa: PLC0415
--- a/home/node_preview.py
+++ b/home/node_preview.py
@@ -10,22 +10,11 @@ AWB_UNAVAILABLE_MESSAGE = (
 )
 
 
-def _load_awb_viewer():
-    from aiidalab_widgets_base import viewer  # noqa: PLC0415
-
-    return viewer
-
-
-def render_unavailable_preview() -> str:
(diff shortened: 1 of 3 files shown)


home/node_preview.py
edan-bainglass, line 1: On first glance, it is unclear why we separate these tiny things into variables/functions. For example, `render_unavailable_preview` is essentially an alias for `AWB_UNAVAILABLE_MESSAGE`. Do we need all these to be separate? If there's a good reason (maybe beyond testing convenience), please state it here. Otherwise, I would merge a few things here (again, unless there's a good reason).

edan-bainglass (commented):
See comment 🙂

danielhollas (approved):
As an alternative, instead of displaying the message, you could display a fallback view of the node with just it's `__repr__` or `__str__`, i.e. the same thing that the AWB viewer does for nodes that don't have dedicated viewer. 

(presumably, if the plan still to pre-install an AWB version to the image in the future, this fallback should be rare)

tests/test_process.py
```diff
@@ -38,16 +63,30 @@ def test_process_inputs_widget(generate_calc_job_node, mock_viewer_module):
...
     widget = home_process.ProcessOutputsWidget(process=multiply_add_completed_workchain)
     widget.show_selected_output(change={"new": "result"})
 
     selected_output = multiply_add_completed_workchain.outputs["result"]
     assert widget.info.value == f"PK: {selected_output.pk}"
+    assert capture_display == [node_preview.render_unavailable_preview()]
+
+
+def test_process_module_does_not_import_awb_directly():
+    assert "aiidalab_widgets_base" not in inspect.getsource(home_process)
```
danielhollas: Not a huge fan of this test, let's remove it.

A proper way to test this would be to import `home.process` in a subprocess and check `sys.modules` for presence of awb. But let's not overcomplicate this.

home/node_preview.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def render_node_preview(node: orm.Node) -> object:
+    """Render a node preview using AWB when it is available."""
+    try:
+        viewer = _load_awb_viewer()
+    except ImportError:
+        return render_unavailable_preview()
+
+    return viewer(node)
```
danielhollas: Given that the plan is (I think) to pre-install AWB in the image anyway (aiidalab/aiidalab-docker-stack#483), I'd simplify all of this and instead if somehow AWB is not available, simply pass the `node` directly into the `display` method, which should just show it's `__repr__` or something like that.

```suggestion
def render_node_preview(node: orm.Node) -> object:
    """Render a node preview using AWB when it is available."""
    try:
        from aiidalab_widgets_base import viewer
    except ImportError:
        return node
    else:
        return viewer(node)
```

danielhollas (changes requested):
@yakutovicha can we get this one merged? Then we can make a `aiidalab-home` release.

home/node_preview.py
yakutovicha: done in [8bad9eb](https://github.com/aiidalab/aiidalab-home/pull/220/commits/8bad9eb072010f506c620c3da6c031259cce476e)

tests/test_process.py
yakutovicha: removed in [4f4880a](https://github.com/aiidalab/aiidalab-home/pull/220/commits/4f4880a4351cfc509e24703df6ebce242643c420)

home/node_preview.py
```diff
@@ -0,0 +1,20 @@
...
+    "`pip install aiidalab-widgets-base` to use this feature."
+)
+
+
+def render_node_preview(node: orm.Node) -> object:
+    """Render a node preview using AWB when it is available."""
+    try:
+        from aiidalab_widgets_base import viewer  # noqa: PLC0415
+    except ImportError:
+        return AWB_UNAVAILABLE_MESSAGE
```
danielhollas: My suggestion here was to return the `node` instead of the AWB_UNAVAILABLE_MESSAGE to simplify things, since AWB should anyway be present in typical case. @yakutovicha  If you disagree can you explain why?

EDIT: Sorry, just saw your message on Slack, I'll defer to Edan what he thinks.

edan-bainglass (commented):
I'm a bit unclear on the goal of this change. AWB import was already guarded with a try/except. This PR appears to simply consolidate it (which is fine/cleaner). But then I would state that more clearly, rather than stating that THIS PR makes AWB a soft dependency. Feels like I'm missing something 😅

home/node_preview.py
edan-bainglass: Unclear how "since AWB should anyway be present in typical case" justifies "return the `node` instead of the `AWB_UNAVAILABLE_MESSAGE`" 🤔 Can you clarify?

In any case, I would do here what is best for the user. If we return the node without an available viewer, the user would just see the `repr` of the node, right? And no clear message as to why they don't see the viewer. Is this useful to the common user? It might be sufficient to some to see the `node` `repr` message, but I would still notify them somehow that there's more to see if they install AWB.

yakutovicha:
> I'm a bit unclear on the goal of this change. AWB import was already guarded with a try/except. This PR appears to simply consolidate it (which is fine/cleaner). But then I would state that more clearly, rather than stating that THIS PR makes AWB a soft dependency. Feels like I'm missing something 😅

Fair point, it is indeed more about consolidation. I am going to rename the PR.

home/node_preview.py
yakutovicha: How about this?

image

danielhollas: Yeah, that looks good to me, thanks! Maybe wrap the command in `<pre>` tags.

edan-bainglass: "For a richer node view, please install the AiiDAlab widgets with `pip install aiidalab-widgets-base`"

Not sure if "node view" is clear to the common user. Maybe "output view" is better.

yakutovicha: > Yeah, that looks good to me, thanks! Maybe wrap the command in `\<pre\>` tags.

`<pre>` puts it on a separate line, so I used `<code>`.

image

here [91e78a7](https://github.com/aiidalab/aiidalab-home/pull/220/commits/91e78a7d1a1db0681f073b01f946e427efb5741d).

yakutovicha: > "For a richer node view, please install the AiiDAlab widgets with `pip install aiidalab-widgets-base`"
> 
> Not sure if "node view" is clear to the common user. Maybe "output view" is better.

done in [3165788](https://github.com/aiidalab/aiidalab-home/pull/220/commits/3165788aede982afad46497d16c4490607b75774).

image

edan-bainglass (approved):
Thanks @yakutovicha. LGTM!

edan-bainglass (approved):

edan-bainglass:
@danielhollas could you also approve this one please? 🙏

danielhollas:
Yrp, will do do after lunch

home/node_preview.py
```diff
@@ -0,0 +1,22 @@
...
+import ipywidgets as ipw
+from aiida import orm
+
+AWB_UNAVAILABLE_MESSAGE = (
+    "For a richer output view, please install the AiiDAlab widgets with "
+    "<code>pip install aiidalab-widgets-base</code>."
+)
+
+
+def render_node_preview(node: orm.Node) -> object:
```
danielhollas: optional: We currently don't type check this code base so I'd prefer to not have the annotations, because it's very easy for them to be wrong (in general) when they're not checked:

```suggestion
def render_node_preview(node):
```

danielhollas (approved):
Thanks @yakutovicha!

PR #222, opened 2026-03-16
Get rid of pandas dependency

### LINKED ISSUE #221: Get rid of pandas dependency
aiidalab-home started to depend on pandas very recently. However, the library is only used to render some tables, something easy enough that we could do ourselves.

yakutovicha:
fixes #221

### DIFF
--- a/home/process.py
+++ b/home/process.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import inspect
 import os
+import re
 import sys
 import threading
 import time
@@ -13,7 +14,6 @@ import warnings
 from collections.abc import Mapping
 
 import ipywidgets as ipw
-import pandas as pd
 import traitlets as tl
 
 # AiiDA imports
@@ -27,6 +27,7 @@ from aiida.cmdline.utils.common import (
 from aiida.common.links import LinkType
 from aiida.tools.query.calculation import CalculationQueryBuilder
 from IPython.display import HTML, Javascript, clear_output, display
+from jinja2 import Template
 
 
 class CantRegisterCallbackError(Exception):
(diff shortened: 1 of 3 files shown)


danielhollas:
Same comment as in the corresponding AWB PR, can we try using tabulate package so we don't have to write HTML and CSS by hand here?

edan-bainglass (approved):
LGTM! @danielhollas `tabulate` seems limited to text-based, very simple tables. I think here, this is fine, for now. Let's not hold it back.

We should, however, settle on a robust table package and use throughout our ecosystem. Another option is to write our own, as @superstar54 did with the `TableWidget` in the QE app. But whatever we do, we should standardize, generalize, and deploy under AWB for use by all!

danielhollas (approved):
>  @danielhollas tabulate seems limited to text-based, very simple tables. 

Can you clarify what you mean by "text based"? Seems like our tables here are pretty simple and we don't need anything fancy?

Okay, I don't want to block, but would be good to write things down in an issue.

@yakutovicha can you merge main and re-run pre-commit. I think this should just re-format the code and fix the failing pre-commit. Alternatively, just merge this and I'll deal with the failing pre-commit in a follow-up PR (sorry about that, I shouldn't have merged that PR).

edan-bainglass:
> Can you clarify what you mean by "text based"? Seems like our tables here are pretty simple and we don't need anything fancy?

Just simply, text-based HTML tables. But yeah, if that's what is used here, one would imagine `tabluate` might work. Since you've raised this issue, mind opening it formally? 🙏

@yakutovicha same comment as [here](https://github.com/aiidalab/aiidalab-widgets-base/pull/737#issuecomment-4141764297)

yakutovicha:
> @yakutovicha same comment as [here](https://github.com/aiidalab/aiidalab-widgets-base/pull/737#issuecomment-4141764297)

Here we go:

before:
image

after:
image


The only noticeable change is the process status. Before, `None` meant "None"; now `None` gets converted to an empty string

yakutovicha:
So, can I merge, @edan-bainglass?

yakutovicha:
> Since you've raised this issue, mind opening it formally?

done https://github.com/aiidalab/aiidalab-home/issues/224

PR #231, opened 2026-07-15
Update a few things for compatibility and styling

edan-bainglass:
See commit messages for details.


danielhollas:
> The `titles` argument is a newer feature. We can switch to it once we
are fully migrated to ipywidgets 8.0.

In `aiidalab-home` we're now fully migrated so I don't think it's necessary?

edan-bainglass:
> > The `titles` argument is a newer feature. We can switch to it once we
> > are fully migrated to ipywidgets 8.0.
> 
> In `aiidalab-home` we're now fully migrated so I don't think it's necessary?

Right, I'm using the latest QE image, which is based on full stack 2026.1031 (pre-upgrades). Okay, that part can be removed. But as @yakutovicha mentioned somewhere, his PR is only a draft and will be split into smaller PRs. Just take from here what you need, if any, then we close.

yakutovicha:
I suggest we do those little changes when we are reviewing specific sections in the follow-up PRs. Is it OK to close this PR, @edan-bainglass?

edan-bainglass:
> I suggest we do those little changes when we are reviewing specific sections in the follow-up PRs. Is it OK to close this PR, @edan-bainglass?

Agreed

PR #232, opened 2026-07-16
Control page: home page icon and panel placeholders

### LINKED ISSUE #155: Implement AiiDA-control page

yakutovicha:
fixes #155 

Basic parts of the control section, extracted from #156.

Introduce a new "AiiDAlab control" page: a home-page entry point for
managing the AiiDA/AiiDAlab installation, with tab-based sections for
status, resources, storage, daemon, processes, profiles, and a danger
zone. Section bodies are placeholders in this PR; real logic for each
section lands in follow-up PRs.

- `start.py`: add a "Control" icon tile linking to `control.ipynb`
- `control.ipynb`: new notebook entry point with a lazily-loaded tab per
  section; each tab probes fresh data both on first open and on every
  revisit, and shows a retryable inline error if construction fails
- `home/control.py`: add `ControlSectionWidget`, the shared base class for
  every section , plus placeholder subclasses for all seven sections
- `tests/test_control.py`: `cover refresh()`'s re-entry guard, success
  path, and error path


control.ipynb
```diff
@@ -0,0 +1,136 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import re\n",
+    "from pathlib import Path\n",
+    "\n",
+    "import ipywidgets as ipw\n",
+    "from aiida import load_profile\n",
+    "\n",
+    "load_profile()\n",
```
edan-bainglass: As a consequence of aiida-core upgrade (I believe due to greenlet, @khsrali?), this part needs to be in its own cell!

home/control.py
```diff
@@ -0,0 +1,147 @@
...
+        if self.description:
+            header_children.append(
+                ipw.HTML(
+                    f"<div style='color:{Theme.COLORS.GRAY};font-size:13px;"
+                    f"margin:2px 0 6px 0;'>{html.escape(self.description)}</div>"
+                )
+            )
+
+        self.refresh_button = ipw.Button(description="Refresh", icon="refresh")
+        self.refresh_button.on_click(self.refresh)
```
edan-bainglass: Let's make this optional (via some input argument), as it may not make sense in some future panels.

edan-bainglass (changes requested):
Made some comments

control.ipynb
danielhollas: More specifically from AiiDAs changelog: https://github.com/aiidateam/aiida-core/blob/main/CHANGELOG.md#replaced-nest_asyncio-with-greenback

> Important change for Jupyter notebook users: When using AiiDA in a Jupyter notebook, load_profile() must now be called in a separate cell before running engine processes:

I think here we'd be fine as we're not gonna run any processes from the Control page, but it also doesn't hurt to move it as a separate line (and before the `home.control` import)

control.ipynb
```diff
@@ -0,0 +1,136 @@
...
+    "    if widget is None:\n",
+    "        try:\n",
+    "            widget = sections[name]()\n",
+    "        except Exception as exc:\n",
+    "            # Show the error but do NOT cache it: revisiting the tab retries.\n",
+    "            placeholder.children = [\n",
+    "                ipw.HTML(\n",
+    "                    f'<div class=\"alert alert-danger\" role=\"alert\">'\n",
+    "                    f'Failed to open the \"{name}\" section: {exc}</div>'\n",
+    "                )\n",
```
danielhollas, line 94: I don't understand why are we catching a broad Exception here??
This is just a dict access right? So KeyError would indicate a programmer error here so not worth catching imo, unless I am missing something

```suggestion
    "        widget = sections[name]()\n",
```

home/control.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def show_plain(self, text):
+        self.info.value = html.escape(text)
+
+    def refresh(self, _=None):
+        if self._refreshing:
+            return
+        self._refreshing = True
+        self.info.value = "Refreshing... <i class='fa fa-spinner fa-spin'></i>"
+        self.refresh_button.disabled = True
```
danielhollas: Probably the button should be disabled first

```suggestion
        self.refresh_button.disabled = True
        self.info.value = "Refreshing... <i class='fa fa-spinner fa-spin'></i>"
```

home/control.py
```diff
@@ -0,0 +1,147 @@
...
+    def show_error(self, text):
+        self.info.value = _state_span("error", text)
+
+    def show_plain(self, text):
+        self.info.value = html.escape(text)
+
+    def refresh(self, _=None):
+        if self._refreshing:
+            return
+        self._refreshing = True
```
danielhollas, line 75:
```suggestion
        if self._refresh_lock.locked():
            return
        with self._refresh_lock:
            ...
```

home/control.py
```diff
@@ -0,0 +1,147 @@
...
+
+    The footer (refresh button, "Last updated" label, transient feedback) and
+    the threaded refresh-with-guard pattern are common to every section, so
+    they live here; subclasses only provide a body and a `_do_refresh()`.
+    """
+
+    description = ""
+
+    def __init__(self, children):
+        self._refreshing = False
```
danielhollas, line 31: I *think* this should be a lock, see below

```suggestion
        self._refresh_lock = threading.Lock()
```

danielhollas:
Thanks for working on this @yakutovicha! Left some comments, most notably regarding thread safety of the refresh method.

control.ipynb
khsrali: > I think here we'd be fine as we're not gonna run any processes from the Control page, but it also doesn't hurt to move it as a separate line (and before the home.control import)

Yes, true!

yakutovicha: fixed in [b3b6174](https://github.com/aiidalab/aiidalab-home/pull/232/commits/b3b61745bc7731ae01c2e0aaa26a3531b9426726)

home/control.py
yakutovicha: done in [01af425](https://github.com/aiidalab/aiidalab-home/pull/232/commits/01af4257c1a0c161997c39ff9c315c318d4add25)

control.ipynb
yakutovicha: It is not about dict access; it is about the widget initialization _call_. Widgets like `DaemonControlWidget`, `StorageWidget`, and others that check the environment might raise exceptions for various reasons: a misconfigured profile, the filesystem not being mounted, or a subprocess call failing.

Without this catch, we might end up with a page in a broken state with no way to recover except reloading the notebook.

home/control.py
yakutovicha: done in [28ffa6c](https://github.com/aiidalab/aiidalab-home/pull/232/commits/28ffa6c734550599a2a5c81279d851ad0c6b7915)

yakutovicha:
> Thanks for working on this @yakutovicha! Left some comments, most notably regarding thread safety of the refresh method.

It seems that the locking mechanism you propose makes things slightly more fragile.

A few observations:

1. The `refresh()` method can't be run more than once at the same time, since everything that calls it lives inside the same thread (Jupyter kernel).
2. For the lock to guard the execution, it needs to live inside the worker:
```python
def refresh(self, _=None):
    if self._refresh_lock.locked():
        return
    def worker():
        with self._refresh_lock:
            self._do_refresh()
    threading.Thread(target=worker, daemon=True).start()
```
3. The worker, in turn, is only launched when the thread actually gets CPU time, so there might be delays (can be milliseconds).

This leaves us with a possibility that two fast clicks can spawn 2 workers at the same time, which the current mechanism prevents entirely.

home/control.py
```diff
@@ -0,0 +1,157 @@
...
+                # Re-enable the button before touching anything else: if a
+                # later step raises, the page must not be left with the
+                # button permanently disabled.
+                if self.refresh_button is not None:
+                    self.refresh_button.disabled = False
+                self._refreshing = False
+                if self._last_updated is not None:
+                    self._last_updated.value = (
+                        f"Last updated: {datetime.now().strftime('%H:%M:%S')}"
+                    )
```
danielhollas: Claude pointed out that this part probably belongs to the `else` branch, since displaying "Last updated" message when a refresh fails is misleading.

home/control.py
```diff
@@ -0,0 +1,157 @@
...
+                # Silent success: the "Last updated" timestamp below already
+                # signals success, so no "refreshed" message is shown.
+                self.info.value = ""
+            finally:
+                # Re-enable the button before touching anything else: if a
+                # later step raises, the page must not be left with the
+                # button permanently disabled.
+                if self.refresh_button is not None:
+                    self.refresh_button.disabled = False
+                self._refreshing = False
```
danielhollas: Minor threading bug pointed out by Claude, these statements should be reversed, otherwise we create a tiny window where the button is enabled but `self._refreshing` is still `True`, and so super-fast click would result in a silent no-op.

```suggestion
                self._refreshing = False
                if self.refresh_button is not None:
                    self.refresh_button.disabled = False
```

edan-bainglass (approved):
Thanks @yakutovicha. LGTM w.r.t my comments.

control.ipynb
danielhollas, line 1: Can you please add the following snippet as a first cell to prevent the dialog box when closing the Tab?
```python
    %javascript
    if (document.getElementById('appmode-busy')) {,
        window.onbeforeunload = function() {return}
    }
```

See #141 for details


image

control.ipynb
```diff
@@ -0,0 +1,146 @@
...
+    "    )\n",
+    "    header = ipw.HTML(\n",
+    "        \"<div style='display:flex;align-items:center;gap:16px;margin:4px 0 12px 0;'>\"\n",
+    "        f\"<div style='width:56px;flex:none;'>{_logo_svg}</div>\"\n",
+    "        \"<div>\"\n",
+    "        \"<h1 style='margin:0;font-size:26px;'>AiiDAlab control</h1>\"\n",
+    "        \"<div style='color:gray;'>Manage your AiiDAlab installation.</div>\"\n",
+    "        \"</div></div>\"\n",
+    "    )\n",
+    "except OSError:\n",
```
danielhollas: nit: I am assuming the the except clause here is guarding if we somehow fail to load the svg? It would be clearer if the other statements were put into the else branch like this

image

Btw: Is this try-except code needed after @edan-bainglass appmode fix?

tests/test_control.py
```diff
@@ -0,0 +1,56 @@
...
+        super().__init__([])
+
+    def _do_refresh(self):
+        self.refresh_calls += 1
+        if self.fail:
+            raise RuntimeError("boom")
+
+
+@pytest.fixture
+def run_threads_synchronously(monkeypatch):
```
danielhollas, line 21: nit: This fixture feels very invasive, it would be nicer to write the tests in a robust way that would use an actual thread. But that can be done in a separate PR.

danielhollas (approved):
Thanks @yakutovicha. I've found a couple other small things (some found by Claude), but happy for you to merge after addressing (or dismissing) them.

I'll reply to the threading issue separately.

danielhollas:
> It seems that the locking mechanism you propose makes things slightly more fragile.

Sorry, my original suggestion indeed did not work, since the lock needs to be released from the worker thread. Here's a patch that I actually tested and seems to work. But I am happy to submit it as a separate PR since this would also require updating the tests and I don't want to block you. 

```diff
diff --git a/home/control.py b/home/control.py
index 19bd92e..7f9e756 100644
--- a/home/control.py
+++ b/home/control.py
@@ -28,7 +28,7 @@ class ControlSectionWidget(ipw.VBox):
     description = ""
 
     def __init__(self, children, show_refresh_button=True):
-        self._refreshing = False
+        self._refresh_lock = threading.Lock()
 
         header_children = []
         if self.description:
@@ -70,9 +70,8 @@ class ControlSectionWidget(ipw.VBox):
         self.info.value = html.escape(text)
 
     def refresh(self, _=None):
-        if self._refreshing:
+        if not self._refresh_lock.acquire(blocking=False):
             return
-        self._refreshing = True
         if self.refresh_button is not None:
             self.refresh_button.disabled = True
         self.info.value = "Refreshing... <i class='fa fa-spinner fa-spin'></i>"
@@ -92,7 +91,7 @@ class ControlSectionWidget(ipw.VBox):
                 # button permanently disabled.
                 if self.refresh_button is not None:
                     self.refresh_button.disabled = False
-                self._refreshing = False
+                self._refresh_lock.release()
                 if self._last_updated is not None:
                     self._last_updated.value = (
                         f"Last updated: {datetime.now().strftime('%H:%M:%S')}"
```

> The refresh() method can't be run more than once at the same time, since everything that calls it lives inside the same thread (Jupyter kernel).

So I actually thought the the kernel uses a separate thread to handle UI interactions, but was apparently wrong about that (and verified by printing `threading.current_thread()` in the refresh method). But that seems a bit like an implementation detail. Interestingly, there's a very recent work in Jupyter ecosystem to enable multiple threads (they call it subshells) so going forward a single-threaded assumption might not hold, but I guess it's fine for now. To enable subshells, one would need to run ipykernel>7.0 (we're running 6.5 now) and jupyterlab>4.4. See following for details (really interesting!)

https://ianthomas23.github.io/jupytercon2025-subshells/#1
https://github.com/ipython/ipykernel/issues/1438

home/control.py
yakutovicha: Agreed, done in [c13a4b3](https://github.com/aiidalab/aiidalab-home/pull/232/commits/c13a4b36df024c703e6109d85e3e219a7204126a).

control.ipynb
yakutovicha: done in [0e801f9](https://github.com/aiidalab/aiidalab-home/pull/232/commits/0e801f979922b935c6328c5bc323ed65638ef37c)

control.ipynb
yakutovicha: agreed, done in [c9c0e41](https://github.com/aiidalab/aiidalab-home/pull/232/commits/c9c0e4133cd68252f07990408536bfbfbded4278)

> Btw: Is this try-except code needed after @edan-bainglass appmode fix?

No, not really. The logo is properly displayed before the fix. It is just to prevent the case where being unable to load the logo kills the whole page.

tests/test_control.py
yakutovicha: Let's do it in a different PR, then. After this base PR is merged, many things can be implemented/discussed in parallel. Do you mind opening issues? 🙏

danielhollas: Opened #235, thanks!

PR #236, opened 2026-07-30
Add `ty` typechecker

### LINKED ISSUE #144: Add type annotations and add mypy as pre-commit check
Same config as in the `aiidalab` package.

danielhollas:
Given the ongoing work on the Control page, I thought it would be helpful to have a typechecker running on the codebase. I chose [ty](https://github.com/astral-sh/ty), the new type-checker from Astral, instead of mypy, for several reasons (roughly in order of priority):

1. It can check jupyter notebooks out of the box (we have a lot of them here!)
2. It is designed to work for untyped codebase, and tries hard to infer types even without type annotations
3. Should generally have fewer false positives
4. Is not a huge pain to configure
5. Is very fast

~~blocked on aiidalab/aiidalab#521~~

LMK what you think! 

Closes #144


__builtins__.pyi
danielhollas, line 1: https://github.com/astral-sh/ruff/pull/22021

.pre-commit-config.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 ci:
     autoupdate_schedule: quarterly
+    skip: [ty]
```
danielhollas, line 4: https://github.com/astral-sh/ty-pre-commit#running-pre-commit-in-ci-with-ty-pre-commit

home/process.py
```diff
@@ -384,7 +384,7 @@ def flatten(d, parent_key="", sep="."):
 
         return options_map
 
-    def show_selected_input(self, change=None):
+    def show_selected_input(self, change):
```
edan-bainglass, line 387: Why drop the default?

edan-bainglass: I guess otherwise ty complained about line 392 and on. But this also means that `show_selected_input` can only be used as a handler, or with an empty dict. I'm guessing we weren't using it directly.

tests/conftest.py
```diff
@@ -74,5 +74,5 @@ def multiply_add_completed_workchain(aiida_local_code_bash):
         "z": orm.Int(3),
         "code": aiida_local_code_bash,
     }
-    _, process = engine.run_get_node(MultiplyAddWorkChain, **inputs)
+    _, process = engine.run_get_node(MultiplyAddWorkChain, **inputs)  # ty: ignore[invalid-argument-type]
```
edan-bainglass, line 77: The `inputs`?

tests/test_computational_resources.py
```diff
@@ -1036,7 +1036,7 @@ def test_optional_new_code_setup_widget():
     assert not hasattr(widget, "btn_setup_new_code")
     assert not hasattr(widget, "_setup_new_code_output")
     assert not hasattr(widget, "resource_setup")
-    selection_row = widget.children[0]  # type: ignore
+    selection_row = widget.children[0]
```
edan-bainglass, line 1039: Is this showing the power of ty? I recall mypy would complain about indexing children. Or did ipywidgets fix their types for children in v8?

tests/test_control.py
```diff
@@ -52,6 +52,6 @@ def test_refresh_shows_error_on_exception(run_threads_synchronously):
     widget = _DummySection(fail=True)
     widget.refresh()
     assert "Failed to refresh" in widget.info.value
-    assert widget.refresh_button.disabled is False
+    assert widget.refresh_button.disabled is False  # ty: ignore[unresolved-attribute]
```
edan-bainglass: Why are these unresolved?

.pre-commit-config.yaml
edan-bainglass, line 4: Skipping ty?

pyproject.toml
```diff
@@ -10,3 +10,12 @@ target-version = "py39"
...
 ignore = ["BLE001"]
+
+# TODO: Fix type errors in these files
+[[tool.ty.overrides]]
+include = [
+  "home/code_setup.py",
+  "home/computational_resources.py",
+]
+[tool.ty.overrides.rules]
+all = "ignore"
```
edan-bainglass: Checking my understanding - do you mean "in these two files, ignore all issues"?

edan-bainglass (commented):
Thanks @danielhollas. Left some comments 🙂

.pre-commit-config.yaml
danielhollas: Yes, because the `pre-commit.ci` service doesn't allow network access which is needed by ty-pre-commit hook to install dependencies. This is explained in https://github.com/astral-sh/ty-pre-commit#running-pre-commit-in-ci-with-ty-pre-commit

danielhollas: We have the same setup in `aiidalab` repo for mypy type checking

pyproject.toml
danielhollas: Indeed, the configuration syntax is weird. But in any case I've removed this and we now typecheck everything

tests/test_control.py
danielhollas: The refresh_button could be None if `show_refresh_button=False`. I've added asserts instead of `ty: ignore`

tests/test_computational_resources.py
danielhollas: Not sure, I think `ty` is just hiding this error somehow, since it thinks that `widget.children` has type Never

```info[revealed-type]: Revealed type
    --> tests/test_computational_resources.py:1039:17
     |
1039 |     reveal_type(widget.children)
     |                 ^^^^^^^^^^^^^^^ `Never`


```

tests/conftest.py
danielhollas: Yeah, `ty` doesn't understand that the `inputs` dict arguments are mapped into `MultiplyAddWorkChain` I think

home/process.py
danielhollas: Indeed, I've checked that we're not calling this function directly anywhere (and `ty` would anyway complain otherwise)

edan-bainglass (approved):
LGTM @danielhollas

danielhollas:
Thanks for the detailed review @edan-bainglass, I'll wait for @yakutovicha approval before merging.

yakutovicha (approved):
The only question I had was "why ty and not mypy?", but I agree that it brings more benefits (looking at the first message in the PR). So go ahead 👍

PR #237, opened 2026-07-31
Update to ruff 0.16.1 and use its default ruleset

danielhollas:
Same as aiidalab/aiidalab#522

This actually fixes two bugs where we were not closing opened files (not using context manager)


tests_notebooks/test_manage_app.py
danielhollas, line 1: These files were previously executable

edan-bainglass (approved):
LGTM! Let's go 💪

PR #239, opened 2026-08-11
Implement `StatusOverviewWidget`

### LINKED ISSUE #233: Control page: implement the Status overview section
Implement `StatusOverviewWidget` to show a compact health table of the services behind AiiDA, one row per item with a colored state icon (ok/warning/error):

- **version** -- the AiiDA version.
- **config** -- the AiiDA config directory path.
- **profile** -- the loaded profile name (error row if no profile is loaded).
- **storage** -- probed with a cheap real query (`QueryBuilder` count), shown as a summary.
- **broker** -- the broker URL with `user:pass@` userinfo stripped, or a warning row if no broker is configured.
- **daemon** -- running/not running with the worker count.

Every probe is individually wrapped, so one failing service renders as an error row instead of taking down the whole table. Refresh via the shared section footer (refresh button + "Last updated"). A working reference can be found in #156.

yakutovicha:
fixes #233


home/control.py
```diff
@@ -116,11 +136,146 @@ def __init__(self):
...
+        return self._status_row("ok", "storage", summary, tooltip=storage_text)
+
+    def _probe_broker(self):
+        broker = manage.get_manager().get_broker()
+        if broker is None:
+            return self._status_row("warning", "broker", "No broker configured")
+        sanitized = _sanitize_broker_url(str(broker))
+        return self._status_row("ok", "broker", sanitized, tooltip=sanitized)
+
+    def _probe_daemon(self):
```
yakutovicha, line 263: @danielhollas, just a small heads-up: since you asked to display information about the workers' CPU utilization.

I intend to add this to the `DaemonControlWidget` -- see the class placeholder above ⬆️.

edan-bainglass:
@yakutovicha I created a new profile (`presto`) and hit refresh. The panel does not reflect the new profile. Changes to the daemon are correctly reflected within a given profile.

yakutovicha:
> @yakutovicha I created a new profile (`presto`) and hit refresh. The panel does not reflect the new profile. Changes to the daemon are correctly reflected within a given profile.

Yes, this is a bit tricky cause AiiDA's `get_profile()` still has the old profile loaded. So technically, the output IS correct, and the old profile is still loaded in the Python kernel.

Maybe a flag message that says something like "Hey, the default profile has changed, please reload the page" would be a good solution to this?

edan-bainglass:
> > @yakutovicha I created a new profile (`presto`) and hit refresh. The panel does not reflect the new profile. Changes to the daemon are correctly reflected within a given profile.
> 
> Yes, this is a bit tricky cause AiiDA's `get_profile()` still has the old profile loaded. So technically, the output IS correct, and the old profile is still loaded in the Python kernel.
> 
> Maybe a flag message that says something like "Hey, the default profile has changed, please reload the page" would be a good solution to this?

I suppose so, if there is no way to do it dynamically. Note that there have been discussions in the aiida-core team regarding "better" handling of profiles. I'll bring this issue up there. Though I'm not sure I fully understand it 😅

yakutovicha:
@edan-bainglass 

image

Changed in https://github.com/aiidalab/aiidalab-home/pull/239/commits/afd0462fb5960bf9604f5e01234954748aad28c0.

edan-bainglass (approved):
@yakutovicha nice! 

I would keep it general though - "Change of profile detected - reload the page to apply". You never know what people name their profiles. For example, suppose you're switching from `alice` to `bob`. Your message would be "alice is now bob - reload the page to apply". Fun, but maybe we avoid this via the general message 🙂

Anyhow, just nitpicking now - feel free to proceed as is.

home/control.py
```diff
@@ -116,11 +137,171 @@ def __init__(self):
...
+
+    Matches through the last `@` so a password containing `@` isn't
+    partially leaked.
+    """
+    return re.sub(r"://[^/ ]+@", "://", url)
+
+
 class StatusOverviewWidget(ControlSectionWidget):
     description = "Health of the services behind AiiDA."
+    _ROW_ICONS: ClassVar[dict[str, str]] = {
```
danielhollas: nit: both _STATE_COLORS and _ROW_ICONS  would be better as an Enum.

home/control.py
```diff
@@ -116,11 +137,171 @@ def __init__(self):
...
+            f"<tr>"
+            f"<td style='padding:1px 8px;'>"
+            f"<span style='color:{color}'>{icon}</span></td>"
+            f"<td style='padding:1px 8px;'><b>{html.escape(label)}</b></td>"
+            f"<td style='color:{text_color};padding:1px 8px;'{title_attr}>"
+            f"{html.escape(text)}</td>"
+            f"</tr>"
+        )
+
+    def _probe_version(self):
```
danielhollas: nit: I find the `probe` naming kind of confusing. These functions ultimately just returns raw html as strings, why not just `_get_version`?
Also we should probably specify that we're talking about AiiDA version here?

```suggestion
    def _get_aiida_version(self):
```

home/control.py
```diff
@@ -116,11 +137,171 @@ def __init__(self):
...
+def _sanitize_broker_url(url) -> str:
+    """Strip credentials (`user:pass@`) from a broker URL, e.g. an AMQP URL.
+
+    Matches through the last `@` so a password containing `@` isn't
+    partially leaked.
+    """
+    return re.sub(r"://[^/ ]+@", "://", url)
+
+
 class StatusOverviewWidget(ControlSectionWidget):
```
danielhollas: Would be good to be more specific with the naming here
```suggestion
class AiidaStatusOverviewWidget(ControlSectionWidget):
```

danielhollas (approved):
Thanks! I don't have time to test this properly this week but trust @edan-bainglass so fine with me to go ahead! I have just some minor nits, feel free to ignore.

Also, should this tab be called "AiiDA status"?

edan-bainglass:
Just a note that I support all of @danielhollas's comments above 🙂

yakutovicha:
> @yakutovicha nice!
> 
> I would keep it general though - "Change of profile detected - reload the page to apply". You never know what people name their profiles. For example, suppose you're switching from `alice` to `bob`. Your message would be "alice is now bob - reload the page to apply". Fun, but maybe we avoid this via the general message 🙂
> 
> Anyhow, just nitpicking now - feel free to proceed as is.

applied in https://github.com/aiidalab/aiidalab-home/pull/239/commits/c14b9211894fea97ebe1dd8c0f059e57fcd10214

home/control.py
yakutovicha: I took this idea a bit further and put both _STATE_COLORS and _ROW_ICONS  under the `State` umbrella. They are ultimately very connected. See [a2f2d8e](https://github.com/aiidalab/aiidalab-home/pull/239/commits/a2f2d8e8ac4e836f931125cc5ab309be4e4d78f3).

home/control.py
yakutovicha: done: [c1b74d8](https://github.com/aiidalab/aiidalab-home/pull/239/commits/c1b74d8d43f2b7a5a90a06c8e77cacead80bc81c)

home/control.py
yakutovicha: done in [5116a04](https://github.com/aiidalab/aiidalab-home/pull/239/commits/5116a04bb1facfb1faa74b9c9842694925352279)

edan-bainglass:
Thanks @yakutovicha. Good to go 👍

